PR #9, opened 2023-08-31
chore: add mac to github workflows

jericht:
We should be building & testing our code on Windows and Mac

Modify GitHub workflows so they are building & testing our code on Windows and Mac. Made similar changes done in https://github.com/casillas2/deadline-cloud/commit/afef801cf3cf2392b53ab36115dc7a4756116a5f#diff-a1a47fe3f012a1f3ed362997e7d812693102beecce28d3eebbd7f446795d84d5

GitHub workflows are building & testing our code on Windows and Mac

GitHub workflows are building & testing our code on Windows and Mac

No

No

🍎 🪟 🍔

### DIFF
--- a/.github/workflows/reuse_python_build.yml
+++ b/.github/workflows/reuse_python_build.yml
@@ -9,13 +9,14 @@ on:
 
 jobs:
   Python:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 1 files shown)


jericht:
Removed Windows CI because adaptor runtime is using Unix sockets which fails the tests on Windows as that's not available

epmog (approved):

amarsjac (approved):

PR #13, opened 2023-09-07
refactor: Update to match some improvements in the Maya submitter

mwiebe:

When updating the Maya submitter, I made improvements that are also useful for Nuke.

* Cleanup to hatch_version_hook.py
* Change the job bundle test runner to match the refactoring done in Maya, including placing the scene file in a scene subdirectory and separating the Nuke-specific parts into a set of interface functions that we can turn into a class in a follow-up refactor.
* 

Easier time using the job bundle output tests.

Ran the job bundle tests.

I ran them before rebase, here are the outputs from that.

```

Running job bundle output test: C:/Users/markw/deadline-clients/deadline-cloud-for-nuke/job_bundle_output_tests\multi-load-save

multi-load-save
Test succeeded

Running job bundle output test: C:/Users/markw/deadline-clients/deadline-cloud-for-nuke/job_bundle_output_tests\noise-saver

noise-saver
Test succeeded

All tests passed, ran 2 total.
```

No

No


epmog (approved):

amznestebanpapp (approved):
LGTM!

PR #14, opened 2023-09-08
fix: job fails if Write node has no folders in filename

amznestebanpapp:

When submitting a job in Nuke that has a `Write` node with a filename that does not contain a folder, the job fails.
Example nk:
```
Root {
 inputs 0
 name colorbars.nk
 frame -3
 format "2048 1556 0 0 2048 1556 1 2K_Super_35(full-ap)"
 proxy_type scale
 proxy_format "1024 778 0 0 1024 778 1 1K_Super_35(full-ap)"
 colorManagement Nuke
 workingSpaceLUT linear
 monitorLut sRGB
 monitorOutLUT rec709
 int8Lut sRGB
 int16Lut sRGB
 logLut Cineon
 floatLut linear
}
ColorBars {
 inputs 0
 barintensity {{curve x1 0 x10 1}}
 name ColorBars1
 xpos -444
 ypos -375
}
set N6b5d800 [stack 0]
Viewer {
 frame -3
 frame_range 1-100
 monitorOutOutputTransform rec709
 name Viewer1
 xpos -444
 ypos -218
}
push $N6b5d800
Write {
 file output_####.exr
 file_type exr
 first_part rgba
 name Write1
 xpos -296
 ypos -259
}
```
Error:
```
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR: Traceback (most recent call last):
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR:   File "/tmp/ojio/session-c3d2e8510f7a4d1c879653acad8b2319rshv_oqi/session-env/packages/nuke/13.2/platform-linux/plugins/nuke_internal/callbacks.py", line 173, in beforeRender
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR:     _doCallbacks(beforeRenders)
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR:   File "/tmp/ojio/session-c3d2e8510f7a4d1c879653acad8b2319rshv_oqi/session-env/packages/nuke/13.2/platform-linux/plugins/nuke_internal/callbacks.py", line 46, in _doCallbacks
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR:     f[0](*f[1],**f[2])
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR:   File "/tmp/ojio/session-c3d2e8510f7a4d1c879653acad8b2319rshv_oqi/session-env/packages/deadline_cloud_for_nuke/0.9.0/_v/a/python/deadline/nuke_adaptor/NukeClient/nuke_client.py", line 48, in ensure_output_dir
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR:     os.makedirs(output_dir)
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR:   File "/tmp/ojio/session-c3d2e8510f7a4d1c879653acad8b2319rshv_oqi/session-env/packages/nuke/13.2/platform-linux/lib/python3.7/os.py", line 223, in makedirs
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR:     mkdir(name, mode)
2023/09/06 17:18:10-07:00 ADAPTOR_OUTPUT: STDERR: FileNotFoundError: [Errno 2] No such file or directory: ''
```

Check that there is a folder in the filename before trying to create one.

Despite uncommon in practice, a `Write` node does not necessarily has to contain a folder. A user could indicate an output filename without a folder. 

Above nk script was used to submit jobs. A job with a filename that does not contain a folder was submitted successful with these changes.
Other changes were done to be able to upload wheels (also tested).

The difference is because I am using a newer version of Nuke. No other differences found:
```
Timestamp: 2023-09-07T19:06:49.873239-07:00
Running job bundle output test: D:\ws\deadline-cloud-for-nuke\job_bundle_output_tests\multi-load-save

multi-load-save
Test failed, found differences
--- expected/parameter_values.yaml
+++ test/parameter_values.yaml
@@ -14,6 +14,6 @@
 - name: ProxyMode
   value: 'false'
 - name: NukeVersion
-  value: 13.2v4
+  value: 14.0v5
 - name: RezPackages
   value: nuke-13 deadline_cloud_for_nuke

Timestamp: 2023-09-07T19:18:28.632054-07:00
Running job bundle output test: D:\ws\deadline-cloud-for-nuke\job_bundle_output_tests\noise-saver

noise-saver
Test failed, found differences
--- expected/parameter_values.yaml
+++ test/parameter_values.yaml
@@ -14,6 +14,6 @@
 - name: ProxyMode
   value: 'false'
 - name: NukeVersion
-  value: 13.2v4
+  value: 14.0v5
 - name: RezPackages
   value: nuke-13 deadline_cloud_for_nuke

Failed 2 tests, succeeded 0.
Timestamp: 2023-09-07T19:19:00.837552-07:00
```

No

Yes, this is updating the dependencies to `deadline-cloud` and `openjobio`, its also changing the module name to `openjd` since that is what is being generated.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -57,10 +57,10 @@ farm that uses a service-managed fleet. You'll need to perform the following ste
 your build of the adaptor for the one in the service.
 
 1. Use the development location from the Submitter Development Workflow.
-   You will need to also check out `openjobio` from git if you do
+   You will need to also check out `openjd` from git if you do
    not already have it. Make sure you're running Nuke with `set DEADLINE_ENABLE_DEVELOPER_OPTIONS=true`
    enabled.
-2. Build wheels for `openjobio`, `deadline-cloud` and `deadline-cloud-for-nuke`, place them in a "wheels"
+2. Build wheels for `openjd`, `deadline-cloud` and `deadline-cloud-for-nuke`, place them in a "wheels"
    folder in `deadline-cloud-for-nuke`. A script is provided to do this, just execute from `deadline-cloud-for-nuke`:
    ```
    $ ./scripts/build_wheels.sh
@@ -68,7 +68,7 @@ your build of the adaptor for the one in the service.
    Wheels should been generated in the "wheels" folder:
    ...
    $ ls ./wheels
-   openjobio-<version>-py3-none-any.whl
+   openjd-<version>-py3-none-any.whl
    deadline-cloud-<version>-py3-none-any.whl
    deadline-cloud-for-nuke-<version>-py3-none-any.whl
    ```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -32,17 +32,19 @@ hatch run all:test
 Here are steps you can follow to run and edit the source code within your Nuke installation on Windows. The steps
 will work on Linux or MacOS if you modify the path references as appropriate.
 
-WARNING: This workflow installs additional Python packages into your Nuke's python distribution.
+WARNING: This workflow installs additional Python packages into your Nuke's python distribution. You may need to 
+run the Command Prompt in Administrative mode if your current user does not have permission to write on Nuke's
+site-package folder.
 
 1. Create a development location within which to do your git checkouts. For example `~/deadline-clients`.
    Clone packages from this directory with commands like
-   `git clone git@github.com:casillas2/deadline-nuke.git`. You'll also want the `deadline` repo.
+   `git clone git@github.com:casillas2/deadline-cloud-for-nuke.git`. You'll also want the `deadline-cloud` repo.
 2. Switch to your Nuke directory, like `cd "C:\Program Files\Nuke13.2v4"`.
-3. Run `.\python -m pip install -e C:\Users\<username>\deadline` to install the Amazon Deadline Cloud Client
+3. Run `.\python -m pip install -e C:\Users\<username>\deadline-clients\deadline-cloud` to install the Amazon Deadline Cloud Client
    Library in edit mode.
-4. Run `.\python -m pip install -e C:\Users\<username>\deadline-nuke` to install the Nuke Submtiter
+4. Run `.\python -m pip install -e C:\Users\<username>\deadline-clients\deadline-cloud-for-nuke` to install the Nuke Submtiter
    in edit mode.
-6. Run `set NUKE_PATH=C:\Users\<username>\deadline-clients\deadline-nuke\src` to put the `menu.py`
+6. Run `set NUKE_PATH=C:\Users\<username>\deadline-clients\deadline-cloud-for-nuke\src` to put the `menu.py`
    file in the path Nuke searches for menu extensions.
 7. Run `set DEADLINE_ENABLE_DEVELOPER_OPTIONS=true` to enable the job bundle debugging support.
    This enables a menu item you can use to run the tests from the `job_bundle_output_tests` directory.
@@ -58,20 +60,21 @@ your build of the adaptor for the one in the service.
    You will need to also check out `openjobio` from git if you do
    not already have it. Make sure you're running Nuke with `set DEADLINE_ENABLE_DEVELOPER_OPTIONS=true`
    enabled.
-2. Build wheels for `openjobio`, `deadline` and `deadline-nuke`.
+2. Build wheels for `openjobio`, `deadline-cloud` and `deadline-cloud-for-nuke`.
    ```
    # If you don't have the build package installed already
    $ pip install build
    ...
    $  mkdir wheels; \
       rm wheels/*; \
-      for dir in ../openjobio ../deadline ../deadline-nuke; \
+      for dir in ../openjobio ../deadline-cloud ../deadline-cloud-for-nuke; \
         do python -m build --wheel --outdir ./wheels --skip-dependency-check $dir; \
       done
    ...
    $ ls ./wheels
    openjobio-<version>-py3-none-any.whl
-   deadline_cloud_for_nuke-<version>-py3-none-any.whl  deadline-<version>-py3-none-any.whl
+   deadline-cloud-<version>-py3-none-any.whl
+   deadline-cloud-for-nuke-<version>-py3-none-any.whl
    ```
 3. Open the Nuke integrated submitter, and in the Job-Specific Settings tab, enable the option 'Include Adaptor Wheels'. This
    option is only visible when the environment variable `DEADLINE_ENABLE_DEVELOPER_OPTIONS` is set to `true`.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -60,16 +60,12 @@ your build of the adaptor for the one in the service.
    You will need to also check out `openjobio` from git if you do
    not already have it. Make sure you're running Nuke with `set DEADLINE_ENABLE_DEVELOPER_OPTIONS=true`
    enabled.
-2. Build wheels for `openjobio`, `deadline-cloud` and `deadline-cloud-for-nuke`.
+2. Build wheels for `openjobio`, `deadline-cloud` and `deadline-cloud-for-nuke`, place them in a "wheels"
+   folder in `deadline-cloud-for-nuke`. A script is provided to do this, just execute from `deadline-cloud-for-nuke`:
    ```
-   # If you don't have the build package installed already
-   $ pip install build
-   ...
-   $  mkdir wheels; \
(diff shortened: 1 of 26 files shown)


src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -203,7 +203,11 @@ def _get_parameter_values(settings: RenderSubmitterUISettings) -> dict[str, Any]
 
     # Set the RezPackages parameter default
     if settings.override_rez_packages:
-        parameter_values.append({"name": "RezPackages", "value": settings.rez_packages})
+        rez_packages = settings.rez_packages
+        # If the adaptor wheels are included, remove the deadline_cloud_for_nuke rez package
+        if settings.include_adaptor_wheels:
+            rez_packages = rez_packages.replace("deadline_cloud_for_nuke", "").strip()
+        parameter_values.append({"name": "RezPackages", "value": rez_packages})
```
mwiebe, line 216: I think this needs logic when override_rez_packages is off as well. Then it would take the default value from the job template as the starting point.

src/deadline/nuke_adaptor/NukeClient/nuke_client.py
```diff
@@ -44,7 +44,8 @@ def __init__(self, socket_path: str) -> None:
         def ensure_output_dir():
             """Ensures the output directory exists before rendering"""
             output_dir = os.path.dirname(nuke.filename(nuke.thisNode()))
-            if not os.path.isdir(output_dir):
+            # Filenames can contain folders, if they do and the folders do not exist, create them
```
mwiebe, line 50: If they do not contain folders, where is the file written? The code that constructs the FlatAssetReferences output directories list needs to know this for correct job attachments behavior. We should add this case to one of our job bundle output test nuke scripts.

pyproject.toml
```diff
@@ -10,8 +10,8 @@ license = ""
 requires-python = ">=3.7"
 
 dependencies = [
-    "deadline == 0.20.*",
-    "openjobio == 0.8.*",
+    "deadline == 0.21.*",
+    "openjd == 0.10.*",
```
mwiebe: We need to wait for stable closed-beta installers before we merge this dep update.

mwiebe (changes requested):

DEVELOPMENT.md
```diff
@@ -58,20 +60,21 @@ your build of the adaptor for the one in the service.
...
    ```
    # If you don't have the build package installed already
    $ pip install build
    ...
    $  mkdir wheels; \
       rm wheels/*; \
-      for dir in ../openjobio ../deadline ../deadline-nuke; \
+      for dir in ../openjobio ../deadline-cloud ../deadline-cloud-for-nuke; \
         do python -m build --wheel --outdir ./wheels --skip-dependency-check $dir; \
       done
```
mwiebe: Since you added the build_wheels.sh script, I think this should just call `./scripts/build_wheels.sh` now.

scripts/build_wheels.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
```
mwiebe, line 3: As a good habit, I like it when we add `set -euo pipefile` to the start of every bash script we write.

src/deadline/nuke_adaptor/NukeClient/nuke_client.py
agerveshi: I believe the `project_directory` if set, otherwise the directory that Nuke was launched from.

amznestebanpapp: This may be reason why I cannot download the results. Let me figure out how to get an absolute path here. Either using `project_directory` as @agerveshi is suggesting or using `cwd` if that does not exist. 
I will also test what it does locally in this scenario where `project_directory` is not set.There is also a "Script Directory" that creates an expression to set the project directory to the one where the script is.

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
amznestebanpapp: ok, I will pass a default value similar to what you did in the maya submitter and then do another PR to replicate your changes.

pyproject.toml
amznestebanpapp: ok, I will revert these changes for the time being

scripts/build_wheels.sh
amznestebanpapp: ok, added it and also added it to `add_copyright_headers.sh` where I copied the header from.

src/deadline/nuke_adaptor/NukeClient/nuke_client.py
amznestebanpapp: looks like `project_directory` is considered as part of `nuke.filename` and it even does the expansion when its the script directory:
```
2023/09/08 11:43:07-07:00 ADAPTOR_OUTPUT: STDOUT: output_filename = /tmp/ojio/session-967eca15f6de4c78a4a23285f4f05f7e5t4580s7/assetroot-ca41c57d5028a6a71bf7/output_####.exr
```
So, will just check if it is not abs, if it is not, will make it relative to `cwd`

scripts/build_wheels.sh
mwiebe: Awesome

src/deadline/nuke_submitter/job_bundle_output_test_runner.py
```diff
@@ -228,6 +228,7 @@ def _run_job_bundle_output_test(test_dir: str, dcc_scene_file: str, report_fh, m
             )
             contents = contents.replace(tempdir, "/normalized/job/bundle/dir")
             contents = contents.replace(tempdir.replace("\\", "/"), "/normalized/job/bundle/dir")
+            contents = contents.replace(os.getcwd(), "/normalized/cwd")
```
mwiebe, line 231: I suspect we should explicitly set the CWD to a known temporary directory different from the job bundle dir, to ensure this has predictable behavior.

mwiebe (approved):

src/deadline/nuke_submitter/job_bundle_output_test_runner.py
amznestebanpapp: Not sure I am following the comment.

the CWD in my case was no the job bundle dir, but the path where I launch Nuke from (which happened to be the root of this repo). 
This currently works because `os.getcwd` does not change from the moment job attachments collects the paths to this point.

Can you clarify what is the unpredictable behavior you are concerned about?

mwiebe: The CWD depends on how the user launched Nuke. It might contain an arbitrary set of files in it, that could be read or ignored by Nuke or scripts embedded in a Nuke scene. Creating a fresh temporary directory, and setting the CWD to it before running the tests eliminates that potential variability.

rmv (approved):
LGTM!

src/deadline/nuke_submitter/assets.py
```diff
@@ -26,7 +26,10 @@ def get_nuke_script_file() -> str:
 
 def get_project_path() -> str:
     """This is the path Nuke uses for relative paths"""
-    return nuke.root().knob("project_directory").getEvaluatedValue()
+    project_path = nuke.root().knob("project_directory").getEvaluatedValue()
+    if not project_path:
+        project_path = os.getcwd()
```
jlangmann-aws, line 31: Does it make sense to set this to the directory of the `nuke.scriptName()` instead of the directory from which Nuke was launched?

amznestebanpapp: Nuke takes cwd as that path. Meaning, if we didnt define `project_directory` and have a relative file path in a `Write` node. The file will go to cwd. 
So here we have to take the same approach, otherwise you could have `project_directory` undefined and a relative path in a `Read` node and assume that you are relative to `nuke.scriptName`, but Nuke is realtive to cwd.
I have seen some cases where TDs set this up relative to the cwd and have artists launch the tool from a specific path (e.g. from within the "shot" path).

agerveshi (approved):

mwiebe (approved):

PR #16, opened 2023-09-08
feat: Add custom OCIO config to job attachments when enabled

agerveshi:
When auto-detecting job attachments, the script file from the root node and then all file knobs from the other nodes in the graph are considered. Additionally, a custom OCIO config file can be specified on the root node. When enabled, this config file and it's corresponding LUTs should also be added as job attachments.

When auto-detecting job attachments, determine if the script is configured to use a custom OCIO config. Load the config in to get the LUT search path. Add the config file and the LUTs directory as job attachments.

Users will be able to use custom OCIO configs without having to manually add the config file and LUTs directory as job attachments.

- Verified the config file and LUTs dir show up in the submitter dialog only when custom OCIO config is enabled. 
- Successfully submitted a job to the farm.
- Added unit test and job bundle output test.

Yes, includes result of added test.

```
Timestamp: 2023-09-19T17:29:17.379558+00:00

multi-load-save
Test succeeded

Timestamp: 2023-09-19T17:29:17.583315+00:00

noise-saver
Test succeeded

Timestamp: 2023-09-19T17:29:17.776855+00:00

ocio
Test succeeded

Timestamp: 2023-09-19T17:29:17.986953+00:00

cwd-path
Test succeeded

All tests passed, ran 4 total.
Timestamp: 2023-09-19T17:29:20.526923+00:00
```

No

No, this is an additive change.

### DIFF
--- a/job_bundle_output_tests/ocio/scene/config.ocio
+++ b/job_bundle_output_tests/ocio/scene/config.ocio
@@ -1,0 +1,25 @@
+ocio_profile_version: 2.1
+
+environment:
+  {}
+search_path:
+  - luts
+  - /tmp/luts
+strictparsing: true
+luma: [0.2126, 0.7152, 0.0722]
+name: deadline-cloud-for-nuke-test-config
+
+roles:
+  {}
+
+file_rules:
+  - !<Rule> {name: Default, colorspace: default}
+
+displays:
+  {}
+
+active_displays: []
+active_views: []
+
+colorspaces:
+  []
--- a/src/deadline/nuke_submitter/assets.py
+++ b/src/deadline/nuke_submitter/assets.py
@@ -7,7 +7,6 @@ import re
 from collections.abc import Generator
 from os.path import commonpath, dirname, join, normpath, samefile
 from sys import platform
-from typing import List
 
 import nuke
 import PyOpenColorIO as OCIO
@@ -47,7 +46,7 @@ def get_custom_ocio_config_path() -> str:
     return nuke.root().knob("customOCIOConfigPath").getEvaluatedValue()
 
 
-def get_custom_ocio_config_search_paths(ocio_config_file: str) -> List[str]:
+def get_custom_ocio_config_search_paths(ocio_config_file: str) -> list[str]:
     """Returns the directories containing the LUTs for the provided OCIO config"""
     ocio_config = OCIO.Config.CreateFromFile(ocio_config_file)
 
--- a/src/deadline/nuke_submitter/assets.py
+++ b/src/deadline/nuke_submitter/assets.py
@@ -9,6 +9,7 @@ from os.path import commonpath, dirname, join, normpath, samefile
 from sys import platform
 
 import nuke
+import PyOpenColorIO as OCIO
 
 from deadline.client.job_bundle.submission import AssetReferences
 from deadline.client.exceptions import DeadlineOperationError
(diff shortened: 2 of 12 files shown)


src/deadline/nuke_submitter/assets.py
```diff
@@ -74,6 +98,14 @@ def get_scene_asset_references() -> FlatAssetReferences:
...
                 asset_references.output_directories.add(dirname(filename))
 
+    # if using a custom OCIO config, add the config file and associated LUT directory
+    if is_custom_ocio_config_enabled():
+        ocio_config_path = get_custom_ocio_config_path()
+        ocio_config_luts_dir = get_custom_ocio_config_luts_dir(ocio_config_path)
+
+        asset_references.input_filenames.add(ocio_config_path)
+        asset_references.input_directories.add(ocio_config_luts_dir)
+
```
rmv, line 114: LGTM if this adds all directories inside the path recursively

src/deadline/nuke_submitter/job_bundle_output_test_runner.py
```diff
@@ -168,17 +184,15 @@ def run_render_submitter_job_bundle_output_test():
             report_fh.write("\n")
             if count_failed:
                 report_fh.write(f"Failed {count_failed} tests, succeeded {count_succeeded}.\n")
-                QMessageBox.warning(
-                    mainwin,
-                    "Some Job Bundle Tests Failed",
-                    f"Failed {count_failed} tests, succeeded {count_succeeded}.\nSee the file {test_job_bundle_results_file} for a full report.",
+                nuke.alert(
```
rmv: nice catch

src/deadline/nuke_submitter/assets.py
```diff
@@ -29,6 +30,29 @@ def get_project_path() -> str:
     return nuke.root().knob("project_directory").getEvaluatedValue()
 
 
+def is_custom_ocio_config_enabled() -> bool:
+    """True if the script is using a custom OCIO config"""
+    return (
+        nuke.root().knob("colorManagement").value() == "OCIO"
+        and nuke.root().knob("OCIO_config").value() == "custom"
```
rmv, line 40: There's one tricky thing wih OCIO which is it also uses env vars to define library settings

https://opencolorio.readthedocs.io/en/latest/guides/using_ocio/using_ocio.html

Those will override the knob settings. Maybe there's a specific Nuke call to grab the OCIO env var it's using taking env vars into account

rmv (approved):
LGTM for a first implementation, I would create a follow up ticket regarding OCIO env vars, or fix it here.

src/deadline/nuke_submitter/assets.py
rmv: Another approach is to just forward OCIO* env vars to the worker and let Nuke do its thing.

rmv:
Also, I think you need to squash all the commits and make the comment there follow the "conventional commits" rules (as you did for the PR title)

src/deadline/nuke_submitter/job_bundle_output_test_runner.py
```diff
@@ -168,17 +184,15 @@ def run_render_submitter_job_bundle_output_test():
             report_fh.write("\n")
             if count_failed:
                 report_fh.write(f"Failed {count_failed} tests, succeeded {count_succeeded}.\n")
-                QMessageBox.warning(
```
agerveshi: Using QMessageBox was causing a segfault on Rocky linux (and possibly windows). Replacing with the equivalent Nuke calls seems to have fixed the issue.

src/deadline/nuke_submitter/assets.py
agerveshi: I will file an internal ticket for forwarding OCIO variables

job_bundle_output_tests/ocio/scene/config.ocio
```diff
@@ -0,0 +1,25 @@
+ocio_profile_version: 2.1
+
+environment:
+  {}
+search_path:
+  - luts
+  - /tmp/luts
```
mwiebe, line 7: I see that we're providing both of these as input directories for job attachments. Does Nuke apply the path mapping to these search paths automatically, or do we have to do some special processing?

Also, do any of the LUT file formats contain further references to additional files?

src/deadline/nuke_submitter/assets.py
mwiebe: Adding them as input directory adds the whole directory tree, yes.

mwiebe (approved):

job_bundle_output_tests/ocio/scene/config.ocio
agerveshi: I added both search paths to this config so we could test relative and absolute. In the relative case OCIO will search relative to the config file's directory. Is that what you mean by path mapping?

Good question about the LUT files, let me take a look.

agerveshi: LUT files do not seem to contain references to other files.

PR #17, opened 2023-09-11
fix: Prevent Nuke from opening multiple submitter dialogs

jlangmann-aws:
Issue: Bea-15149
Previously, a user could open multiple submitter dialogs. The user should not be able to open more than 1 submitter dialog.

If the submitter dialog is already open, refresh the dialog properties (including job attachments). 

Better user experience

Job Bundle Output tests were successful

Yes, ran tests. Also tested adding/removing read node dependencies and re-opening the Submitter dialog. 

```
```
only diff was my Nuke version (I'm using 14v5)

This change is dependent on - https://github.com/casillas2/deadline-cloud/pull/35

### DIFF
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -179,6 +179,8 @@ def _get_parameter_values(
 
 
 def show_nuke_render_submitter(parent, f=Qt.WindowFlags()) -> "SubmitJobToDeadlineDialog":
+    global g_submitter_dialog
+
     render_settings = RenderSubmitterUISettings()
 
     # Set the setting defaults that come from the scene
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -26,6 +26,8 @@ from .ui.components.scene_settings_tab import SceneSettingsWidget
 from deadline.client.job_bundle.submission import AssetReferences
 from deadline.client.exceptions import DeadlineOperationError
 
+g_submitter_dialog = None
+
 
 def show_nuke_render_submitter_noargs() -> "SubmitJobToDeadlineDialog":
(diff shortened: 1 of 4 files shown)


src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -262,15 +266,18 @@ def job_bundle_callback(
...
+            SceneSettingsWidget,
+            render_settings,
+            auto_detected_attachments,
+            attachments,
+            job_bundle_callback,
+            parent=parent,
+            f=f,
+        )
+    else:
+        g_submitter_dialog.refresh(auto_detected_attachments, attachments)
```
mwiebe: I think we'll want render_settings to be part of this refresh as well. If we're refreshing an existing dialog, it may be necessary to move the job_bundle_callback out of this function to no longer be a closure. This is because after updating this, the callback is still a closure to the first call that created the dialog, not any of the later updates.

mwiebe (changes requested):

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
jlangmann-aws: Makes sense, happy to do this while I'm at it. It will also be a slightly larger change in `deadline-cloud` since we'll need to add refresh functions for the Shared Job Settings and the Job Specific Settings UIs.

jlangmann-aws: Updated this and the deadline-cloud PR so we're now passing in the render settings. It looks like the `job_bundle_callback` is valid as is since we're passing in the settings (which are reset if necessary) when submitting a job or saving a bundle. 

Also I see that the test coverage went slightly low with python 3.10. I can merge in this PR after Ting's review https://github.com/casillas2/deadline-cloud-for-nuke/pull/18 is complete!

amznestebanpapp (approved):

tincheng (approved):

agerveshi (approved):

mwiebe (approved):

PR #18, opened 2023-09-11
fix: adaptor wheel error dialog and unit test coverage

tincheng:
- Unit test coverage needed improvement
- There was a small bug where if we tried to submit without setting up wheel directory but had "developer options: include adaptor wheels" selected, then it gave a generic `FileNotFoundError` instead of the intended `The Developer Option 'Include Adaptor Wheels' is enabled, but the wheels directory does not exist` message

- increased unit test coverage to ~77%
- changed the check for directory to rely on just `is_dir()` instead of `exists()`

Better test coverage and small fix for user experience

- coverage updated from `hatch run test`
- locally verified that intended Deadline error message was shown in UI instead of FileNotFound error

No - I was not able to run the Job Bundle Output Tests on Mac. Got ```Failed 2 tests, succeeded 0.
See the file /Users/tincheng/workplace/deadline-cloud-for-nuke/job_bundle_output_tests/test-job-bundle-results.txt for a full report.``` and the output file was blank. Probably need more time to get the bundle tests working on Mac, but I think this PR should have low impact without this as changes are either just unit tests or had been verified within DCC. 

```
```

No

No


test/deadline_adaptor_for_nuke/unit/NukeAdaptor/test_adaptor.py
```diff
@@ -575,6 +598,40 @@ def test_handle_progress(
         ("Eddy[ERROR] - Something terrible happened", 3),
     ]
 
+    @pytest.mark.parametrize("regex_index, stdout, expected_progress", handle_progess_params)
```
rmv: typo -> should be handle_progress_params, maybe fix in rest of instances?

test/deadline_adaptor_for_nuke/unit/NukeAdaptor/test_adaptor.py
```diff
@@ -673,3 +730,34 @@ def test_does_nothing_if_nuke_not_running(
         # THEN
         assert "CANCEL REQUESTED" in caplog.text
         assert "Nothing to cancel because Nuke is not running" in caplog.text
+
+
+class TestNukeAdaptor_get_major_minor_version:
```
rmv, line 733: Missing type hints in these methods

agerveshi, line 733: Do we know what actually calls `_get_major_minor_version`? I see that all of the adapters implement it, but what is it used for?

test/deadline_submitter_for_nuke/unit/test_deadline_submitter_for_nuke.py
```diff
@@ -0,0 +1,489 @@
...
+        return_value=TEST_NUKE_SCRIPT_FILE_PATH,
+    ) as m:
+        yield m
+
+
+@pytest.fixture
+def mock_sticky_file_path():
+    with patch(
+        "deadline.nuke_submitter.deadline_submitter_for_nuke._get_sticky_settings_file",
+        return_value=Path("/some/path.deadline_settings.json"),
```
agerveshi: This path is hard-coded a few times, is there a way to get the return value from the fixture in the other locations rather than duplicating it?

test/deadline_submitter_for_nuke/unit/test_template_output/expected_job_template_output.py
```diff
@@ -0,0 +1,192 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+EXPECTED_NUKE_JOB_TEMPLATE = {
```
agerveshi: Maybe add a comment on how you generated this template to make it easier for others to update it in the future?

agerveshi (approved):

test/deadline_adaptor_for_nuke/unit/NukeAdaptor/test_adaptor.py
tincheng: fixed!

test/deadline_adaptor_for_nuke/unit/NukeAdaptor/test_adaptor.py
tincheng: it is not currently used! I saw an internal ticket in the backlog on using this function instead of a hard-coded DCC version, so this function was likely added in preparation for that task.

test/deadline_submitter_for_nuke/unit/test_deadline_submitter_for_nuke.py
tincheng: Sure, I will make it into a constant

test/deadline_submitter_for_nuke/unit/test_template_output/expected_job_template_output.py
tincheng: That's a good question. I generated this template under the assumption that the current implementation was correct. So the templates here used for comparison were simply generated by calls to `_get_job_template()`, then pasting the results. Ideally if people make only small changes in the future, they will not need to update the whole template, and can just fix the small parts that become different. If there are large changes to the whole template, it may be easier to re-generate the template and overriding the existing one and just check diff to make sure the changes look reasonable.

amznestebanpapp: instead of putting the whole output in a str, I would suggest to have it as a file and make the test load the file and extract the contents. There is a vscode plugin to help write these templates, so it is easier to have it as a file directly

amznestebanpapp (approved):

rmv (approved):
LGTM (though I'm new to this testing setup)

test/deadline_submitter_for_nuke/unit/test_template_output/expected_job_template_output.py
tincheng: Changed the files to json in latest commit

tincheng: Changed the files to json in latest commit

jlangmann-aws (approved):

test/deadline_submitter_for_nuke/unit/test_deadline_submitter_for_nuke.py
```diff
@@ -0,0 +1,513 @@
...
+import json
+import os
+from json import JSONDecodeError
+from pathlib import Path
+from unittest.mock import MagicMock, patch, mock_open
+
+import nuke
+
+from deadline.nuke_submitter.deadline_submitter_for_nuke import (
+    _get_sticky_settings_file,
```
mwiebe: We're going to have conflicts to resolve on these ones with https://github.com/casillas2/deadline-cloud-for-nuke/pull/24, as I simplified the sticky setting logic.

test/deadline_submitter_for_nuke/unit/test_deadline_submitter_for_nuke.py
```diff
@@ -0,0 +1,513 @@
...
+    _load_sticky_settings,
+    _save_sticky_settings,
+    get_nuke_version,
+    _get_write_node,
+    _get_job_template,
+    _get_parameter_values,
+)
+from deadline.nuke_submitter.data_classes.submission import (
+    RenderSubmitterSettings,
```
mwiebe: Same here, I removed this dataclass.

test/deadline_submitter_for_nuke/unit/test_deadline_submitter_for_nuke.py
```diff
@@ -0,0 +1,513 @@
...
+TEST_NUKE_SCRIPT_FILE_PATH = "/some/path.nk"
+TEST_NUKE_SCRIPT_SETTINGS_PATH = "/some/path.deadline_settings.json"
+REZ_PACKAGE_DEFAULT = "nuke-13 deadline_cloud_for_nuke"
+
+
+@pytest.fixture
+def expected_nuke_job_template():
+    with open(
+        os.path.join(
+            os.path.dirname(__file__), "test_expected_output_json/expected_job_template_output.json"
```
mwiebe: Hmm, I'd prefer we don't do anything like this in our unit tests. The job bundle output tests serve this purpose, and we should rely exclusively on them instead. In particular, we should add those tests to our CI/CD pipeline.

Why is adding tests like this bad? The testing coverage gains are not flexible for the purpose of the code - they instead lock down the implementation details. So when we do a refactor that keeps the same functionality but dramatically changes the shape of the job templates, these need to be re-generated. The job bundle output tests have a built-in mechanism for that: You delete the folder and re-run the test, it generates the output again, and you validate it by hand. These json files don't work the same way.

Could you remove these tests for now? Maybe we can re-evaluate them, but first I want the job bundle outputs in our CI/CD pipeline. A useful change would be to add coverage analysis to those job bundle output tests. Then we could combine the coverage data from them with the coverage data from the unit tests to evaluate our overall coverage metric.

mwiebe (changes requested):

test/deadline_submitter_for_nuke/unit/test_deadline_submitter_for_nuke.py
tincheng: I have removed these tests for now, and will take a look into job bundle test coverage analysis.

test/deadline_submitter_for_nuke/unit/test_deadline_submitter_for_nuke.py
tincheng: Ah! will rebase again

test/deadline_submitter_for_nuke/unit/test_deadline_submitter_for_nuke.py
tincheng: I saw that load_sticky and save_sticky were moved there, but I won't add new tests for them this time

mwiebe (approved):

PR #27, opened 2023-09-22
fix!: rename install builder component to match other submitters

epmog:

From the deadline cloud submitter installer help menu:
```
 --enable-components <enable-components>     Comma-separated list of components
                                             Default: deadline_client
                                             Allowed: deadline_cloud_for_maya nuke_integrated_submitter
```
note the inconsistencies in the available components

make them match!

consistency!

N/A, it now looks like https://github.com/casillas2/deadline-cloud-for-maya/blob/mainline/install_builder/deadline-cloud-for-maya.xml

N/A

N/A

yep, changed how cli option name for installer

### DIFF
--- a/install_builder/deadline-cloud-for-nuke.xml
+++ b/install_builder/deadline-cloud-for-nuke.xml
@@ -1,7 +1,7 @@
 <component>
-    <name>nuke_integrated_submitter</name>
-    <description>Integrated submitter for Nuke</description>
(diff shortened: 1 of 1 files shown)


jericht (approved):
LGTM, thanks for fixing this!

gahyusuh (approved):

lucaseck (approved):

PR #34, opened 2023-10-05
feat: Apply path mapping for OCIO configs with absolute search paths

agerveshi:
BEA-15636

The search paths in an OCIO config can contain relative and absolute paths. When using path mapping (such as in the case that the OCIO config was uploaded as a job attachment), the the absolute search paths need to be updated for the render to succeed.

In the NukeClient, determine if a custom OCIO config is being used, and if it contains absolute search paths. If so, create a temporary copy of the config where all of the paths are absolute and have had path mapping applied. Since the temporary file is being stored in a temporary directory (in the case that the original config's directory is read-only), the the relative paths will not be valid.

When using custom OCIO configs with absolute search paths, path mapping will be applied properly without the user needing to intervene.

1. Created a modified OCIO config with absolute search paths
2. Copied the config to a second directory, to represent the job attachment version of the config
3. Ran the NukeAdaptor locally with a path mapping rule from the original to copied OCIO directory.
4. Verified that the path mapping was applied in the logs, and that a temporary OCIO config was generated with absolute search paths that match the path mapping rule.

Waiting on an adaptor runtime fix to be able to test with a job submission.

```
Timestamp: 2023-10-05T00:46:50.416902+00:00

multi-load-save
Test succeeded

Timestamp: 2023-10-05T00:46:50.705180+00:00

noise-saver
Test succeeded

Timestamp: 2023-10-05T00:46:50.855053+00:00

ocio
Test succeeded

Timestamp: 2023-10-05T00:46:51.024486+00:00

cwd-path
Test succeeded

All tests passed, ran 4 total.
Timestamp: 2023-10-05T00:46:52.460321+00:00

```

Not the OCIO parts specifically, but updated the development documentation with directions on running the NukeAdaptor locally.

No


agerveshi:
This PR also shuffles the unit tests around so that they are under a single `unit` folder like the other repos. I also fixed the markdown formatting warnings in the DEVELOPMENT.md

pyproject.toml
```diff
@@ -41,6 +41,7 @@ path = "hatch_custom_hook.py"
 [tool.hatch.build.hooks.custom.copy_version_py]
 destinations = [
   "src/deadline/nuke_adaptor",
+  "src/deadline/nuke_ocio",
```
rcajigal: super minor: feels like eventually, moving this under a package grouping more nuke specific tooling/util would make more sense rather than at the top level with the adaptor/submitter

src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
```diff
@@ -410,14 +410,11 @@ def _start_nuke_client(self) -> None:
...
 
-        Args:
-            nuke_version (str): The version of Nuke that we are launching.
-
         Raises:
             FileNotFoundError: If the nuke_client.py file could not be found.
             KeyError: If a configuration for the given platform and version does not exist.
         """
-        nuke_exe = "nuke"
+        nuke_exe = os.environ.get("NUKE_ADAPTOR_NUKE_EXECUTABLE", "nuke")
```
rcajigal, line 417: Is this just for local testing? Or are we expecting it to be set from the queue env

rcajigal (approved):

src/deadline/nuke_ocio/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+
+from .nuke_ocio import (
```
jlangmann-aws: This seems a bit unnecessary here to have imports to all functions from another module... any reason for it? I agree with Ryan's comment above too that it makes sense to move this under a Nuke package group.

src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
amznestebanpapp: this is giving a way to override the executable, if `NUKE_ADAPTOR_NUKE_EXECUTABLE` is not provided, it will use `nuke` which was the previous value.

test/unit/deadline_adaptor_for_nuke/NukeClient/test_client.py
```diff
@@ -269,3 +270,57 @@ def test_recursive_map_path(
...
+                "/session-dir/ocio/luts",
+                "/session-dir/absolute/path/to/luts",
+            ]
+
+            expected_updated_config_path = os.path.join(
+                os.environ["NUKE_TEMP_DIR"],
+                os.path.basename(mock_get_custom_ocio_config_path.return_value),
+            )
+
+            client = NukeClient(socket_path="/tmp/9999")
```
amznestebanpapp: I have seen this running into trouble over time with devs copying-pasting and then trying to run things in parallel. 
I would suggest to use something like `tempfile.TemporaryDirectory()`.

amznestebanpapp (approved):

src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
agerveshi: Yes this gives us a way to override for local testing. If a customer had their own rez package and the executable wasn't called `nuke` then they could set this in their queue environment rather than change their package or modify the source code.

pyproject.toml
agerveshi: Renamed to `nuke.ocio_util`

jlangmann-aws (approved):

PR #39, opened 2023-10-11
fix: check for modified script file and recurse group nodes for asset references

agerveshi:
Bea-16246

- The nuke submitter would allow for a job to be submitted when the script has unsaved changes. This meant that the state of the node graph being submitted doesn't match the artist's current session.
- Traversing the node graph for asset references was not recursing into group nodes, potentially missing input files.

- Check if the node graph has been modified since the last save when launching the submitter dialog
- set the `recurseGroups` flag when getting the nodes to traverse

- Artist's will have to save their script before submitting a job
- automatic job attachments will be more robust

- Attempted to open and the submitter dialog with unsaved changes, which fails. I can open it again once I save
- Put a Read node in a group and saw that the inputs are now in the job attachments UI
- Added a job bundle output test for Read node inside a Group

```
Timestamp: 2023-10-11T23:39:14.135434+00:00

multi-load-save
Test succeeded

Timestamp: 2023-10-11T23:39:14.364478+00:00

noise-saver
Test succeeded

Timestamp: 2023-10-11T23:39:14.516072+00:00

ocio
Test succeeded

Timestamp: 2023-10-11T23:39:14.676970+00:00

cwd-path
Test succeeded

Timestamp: 2023-10-11T23:39:14.821270+00:00
Running job bundle output test: /home/rocky/src/deadline-cloud-for-nuke/job_bundle_output_tests/group-read

group-read
Test succeeded

Timestamp: 2023-10-11T23:39:17.390063+00:00
```

No

No


src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -185,19 +185,24 @@ def _get_parameter_values(
...
-    render_settings.frame_list = str(nuke.root().frameRange())
-    render_settings.is_proxy_mode = nuke.root().proxy()
-
     script_path = get_nuke_script_file()
     if not script_path:
         raise DeadlineOperationError(
             "The Nuke Script is not saved to disk. Please save it before opening the submitter dialog."
         )
 
+    if nuke.root().modified():
```
jlangmann-aws, line 194: Shouldn't this check happen before submitting rather than before opening the dialog? The user could open the dialog from a saved scene, make changes without saving, and then hit "submit". 

Maybe the logic should live in the create_job_bundle_callback?

amznestebanpapp: +1, however, I understand the same can be said on the above check
Looks like the Submit dialog is not modal either... I was able to close the comp with the dialog open. I guess it will still pick up the nk file. Seems like it only needs the script path, and then can do the `modified` check later on... maybe do it on both sides?

amznestebanpapp (approved):

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
jlangmann-aws: Side note -> IMO I think it's good that the Submit dialog is not modal. I could see an artist wanting to refer back to the script while adjusting the job-specific settings (i.e. frame range, write node selection, etc).

agerveshi: The traversal for job attachments also happens when you open the dialog, so the artist could have made additional changes to the script after opening the dialog. What would we do in that case?

rmv: As currently implemented submit buttons SHOULD be modal... but they were not in Thinkbox Deadline, and I agree with Julie that it's best if they are not.

This is typically solved by adding "refresh/apply" buttons or having it happen automatically without losing user input data.

So Alex's change does not make the issue worse and I think it's ok to commit, but I would also delay the test to the submit button as suggested by Julie and Esteban (here or in another ticket).

rmv (approved):
LGTM!

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
agerveshi: Created internal ticket Bea-16288

PR #44, opened 2023-10-18
feat: Adding telemetry events for adaptor and submitter actions

rcajigal:
There currently exists no submitter specific telemetry events for Nuke for neither the submitter nor adaptor. Because of this, there is a lack of visibility into user behavior on the submitter, and a general lack of visibility around the versioning of the DCC a user is running on either their workstation, or as a worker.

Using deadline-cloud's telemetry library to record appropriate events and adding information when needed. Existing information regarding telemetry session-id can be used to aggregate metrics in the future. One caveat that was needed in this solution was the ability to respect telemetry opt-out. The adaptor running on the workers didn't respect the opt-out flag that I thought would have been set by the worker agent from this [PR](https://github.com/casillas2/deadline-cloud-worker-agent), so instead I'm passing in an optional parameter to take it from the workstation. Another possible change is the versioning information for the adaptor as right now, it's simply trying to obtain it from the environment variables passed in from Rez, but this isn't guaranteed and can likely be standardized across all integrations.

There will be metrics for Nuke specifically showing when users open up the submitter UI and when the adaptor is running on a worker and what version(s) their running.

Against the alpha telemetry endpoint. I can see and verify that all the information added is present in CloudWatch.

Ran it and encountered the same error as the last merged PR where a new RezPackages field is getting inserted. Regarding the new field for telemetry opt out, looks fine otherwise. Will maybe take a look after this if I have bandwidth.
```

Timestamp: 2023-10-17T14:15:42.511358-07:00
Running job bundle output test: /Volumes/workplace/bealine-clients/deadline-cloud-for-nuke/job_bundle_output_tests/group-read

group-read
Test succeeded

Timestamp: 2023-10-17T14:15:42.962052-07:00
Running job bundle output test: /Volumes/workplace/bealine-clients/deadline-cloud-for-nuke/job_bundle_output_tests/noise-saver

noise-saver
Test succeeded

Timestamp: 2023-10-17T14:15:43.336982-07:00
Running job bundle output test: /Volumes/workplace/bealine-clients/deadline-cloud-for-nuke/job_bundle_output_tests/multi-load-save

multi-load-save
Test succeeded

Timestamp: 2023-10-17T14:15:43.740735-07:00
Running job bundle output test: /Volumes/workplace/bealine-clients/deadline-cloud-for-nuke/job_bundle_output_tests/ocio

ocio
Test failed, found differences
Missing files: ['test-job-bundle-results.txt']
--- expected/parameter_values.yaml
+++ test/parameter_values.yaml
@@ -3,10 +3,12 @@
   value: 1-100
 - name: NukeScriptFile
   value: /normalized/job/bundle/dir/ocio.nk
+- name: TelemetryOptOut
+  value: 'false'
 - name: ProxyMode
   value: 'false'
-- name: TelemetryOptOut
-  value: 'false'
+- name: RezPackages
+  value: nuke-13 deadline_cloud_for_nuke
 - name: deadline:targetTaskRunStatus
   value: READY
 - name: deadline:maxFailedTasksCount

Timestamp: 2023-10-17T14:15:44.132550-07:00
Running job bundle output test: /Volumes/workplace/bealine-clients/deadline-cloud-for-nuke/job_bundle_output_tests/cwd-path

cwd-path
Test failed, found differences
--- expected/parameter_values.yaml
+++ test/parameter_values.yaml
@@ -3,10 +3,12 @@
   value: 1-100
 - name: NukeScriptFile
   value: /normalized/job/bundle/dir/cwd-path.nk
+- name: TelemetryOptOut
+  value: 'false'
 - name: ProxyMode
   value: 'false'
-- name: TelemetryOptOut
-  value: 'false'
+- name: RezPackages
+  value: nuke-13 deadline_cloud_for_nuke
 - name: deadline:targetTaskRunStatus
   value: READY
 - name: deadline:maxFailedTasksCount

Failed 2 tests, succeeded 3.
Timestamp: 2023-10-17T14:15:51.118660-07:00

```

N/A

No - shouldn't be, deadline-cloud changes have already been merged in and released and the opt-out flag should be optional

### DIFF
--- a/src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
+++ b/src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
@@ -10,6 +10,7 @@ import threading
 import time
 from typing import Callable, cast
 
+from deadline.client.api import get_deadline_cloud_library_telemetry_client, TelemetryClient
 from openjd.adaptor_runtime.adaptors import Adaptor, AdaptorDataValidators
 from openjd.adaptor_runtime_client import Action
 from openjd.adaptor_runtime.process import LoggingSubprocess
@@ -295,6 +296,14 @@ class NukeAdaptor(Adaptor):
         self._server_thread = self._start_nuke_server_thread()
         self._populate_action_queue()
 
+        # initialize telemetry client to handle opt out
+        self._get_deadline_telemetry_client(self.init_data.get("telemetry_opt_out", False))
+        self._record_adaptor_runtime_event(
+            self.__class__.__name__,
+            "on_start",
+            self._get_major_minor_version(os.environ.get("NUKE_VERSION", "")),
+        )
+
         self._start_nuke_client()
 
         is_timed_out = self._get_timer(self._NUKE_START_TIMEOUT_SECONDS)
@@ -322,6 +331,7 @@ class NukeAdaptor(Adaptor):
         self.validators.run_data.validate(run_data)
         self._is_rendering = True
         self._action_queue.enqueue_action(Action("start_render", {"frame": run_data["frame"]}))
+
         while self._nuke_is_running and self._is_rendering and not self._has_exception:
             time.sleep(0.1)  # busy wait so that on_cleanup is not called
 
@@ -329,6 +339,10 @@ class NukeAdaptor(Adaptor):
             #  This is always an error case because the Nuke Client should still be running and
             #  waiting for the next command. If the thread finished, then we cannot continue
             exit_code = self._nuke_client.returncode
+
+            self._get_deadline_telemetry_client().record_error(
+                {"exit_code": exit_code}, str(RuntimeError)
+            )
             raise RuntimeError(
                 "Nuke exited early and did not render successfully, please check render logs. "
                 f"Exit code {exit_code}"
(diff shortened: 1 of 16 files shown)


src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
```diff
@@ -456,3 +470,23 @@ def _populate_action_queue(self) -> None:
         for name in _NUKE_INIT_KEYS:
             if name in self.init_data:
                 self._action_queue.enqueue_action(Action(name, {name: self.init_data[name]}))
+
+    def _get_deadline_telemetry_client(self, adaptor_opt_out: bool = False) -> TelemetryClient:
+        """
+        Wrapper around the Deadline Client Library telemetry client, in order to set package-specific information
+        """
+        client = get_deadline_cloud_library_telemetry_client()
+        client.telemetry_opted_out = client.telemetry_opted_out or adaptor_opt_out
```
tincheng, line 479: I think it's better to set `client.telemetry_opted_out = adaptor_opt_out`? In case client default is True, we still want to honor the adaptor setting? Also a nit but I think the name `telemetry_opted_out` represents the field better than `adaptor_opt_out`, since that's the name from init_data.

rmv: I agree with the code here: if the adaptor or the client opt out, there should be no telemetry data recorded.

rmv (approved):
LGTM

src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
rcajigal: Yea, the ideal case here is that the worker had already set opt-out and this line doesn't have to be here, but that needs some more investigation. This hack-y logic is meant to handle the case that either it was already set (either by a previous initialization) or explicitly set via `adaptor_opt_out` (which should only ever appear here hence the naming). The client defaults to `False`, so if we ever see the opt-out flag appear to be True, imo it's better to always opt-out

tincheng (approved):

PR #46, opened 2023-10-24
feat: [Nuke] Added MOV support

jlangmann-aws:
Previously the nuke submitter was not able to handle rendering "MOV" renders. The submitter could only support single frame renders rather than handle rendering out a frame range in a single task.

The solution adds basic "MOV" render functionality to the submitter. If a "MOV" render is detected via the write node, a single task is generated for the job. 

If MOV is detected:
1. Remove the parameter definition "Frame" from the step
2. Update the script data with the "frameRange" as specified by the frame range.

These template changes result in a single task being created

Add support for MOV rendering!

Yes.
1. Tested rendering out a file sequence (exr) to verify existing functionality
2. Tested rendering out a mov file to verify that a single task was created

From both of the tests above I also verified the generated output in Nuke

```

Timestamp: 2023-10-24T21:31:25.428295+00:00
Running job bundle output test: /home/rocky/git/deadline-cloud-for-nuke/job_bundle_output_tests/multi-load-save

multi-load-save
Test succeeded

Timestamp: 2023-10-24T21:31:25.682390+00:00
Running job bundle output test: /home/rocky/git/deadline-cloud-for-nuke/job_bundle_output_tests/noise-saver

noise-saver
Test succeeded

Timestamp: 2023-10-24T21:31:25.857324+00:00
Running job bundle output test: /home/rocky/git/deadline-cloud-for-nuke/job_bundle_output_tests/cwd-path

cwd-path
Test succeeded

Timestamp: 2023-10-24T21:31:26.048854+00:00
Running job bundle output test: /home/rocky/git/deadline-cloud-for-nuke/job_bundle_output_tests/ocio

ocio
Test succeeded

Timestamp: 2023-10-24T21:31:26.255416+00:00
Running job bundle output test: /home/rocky/git/deadline-cloud-for-nuke/job_bundle_output_tests/group-read

group-read
Test succeeded

Timestamp: 2023-10-24T21:31:28.318128+00:00

```

No

No


src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -121,6 +122,34 @@ def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
...
+            frame_list = (
+                settings.frame_list
+                if settings.override_frame_range
+                else str(write_node.frameRange())
+            )
+            match = re.match(r"(\d+)-(\d+)", frame_list)
+            if not match:
+                raise DeadlineOperationError(
+                    "Invalid frame range {} for evaluating a MOV render. Frame range must follow the format 'startFrame - endFrame'".format(
+                        frame_list
```
amznestebanpapp: you can use f-string formatting here

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -121,6 +122,34 @@ def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
...
+
+            start_frame = match.group(1)
+            end_frame = match.group(2)
+
+            # Remove the Frame parameter space and update the script data with the desired start and end frame
+            for step in job_template["steps"]:
+                del step["parameterSpace"]
+                step["script"]["embeddedFiles"][0]["data"] = "frame: {}\n" "endframe: {}\n".format(
+                    start_frame, end_frame
+                )
```
amznestebanpapp: same here

amznestebanpapp (approved):

src/deadline/nuke_adaptor/NukeClient/nuke_handler.py
```diff
@@ -53,9 +53,11 @@ def start_render(self, data: dict) -> None:
...
             RuntimeError: If start render is called without a frame number.
         """
-        frame = data.get("frame")
-        if frame is None:
-            raise RuntimeError("NukeClient: start_render called without a frame number.")
+        start_frame = data.get("frame")
+        if start_frame is None:
+            raise Exception("NukeClient: start_render called without a frame number.")
+
+        end_frame = data["endframe"] if data.get("endframe") is not None else start_frame
```
mwiebe: nit: Could simplify to:

```python
end_frame = data.get("endframe", start_frame)
```

src/deadline/nuke_adaptor/NukeAdaptor/schemas/run_data.schema.json
```diff
@@ -2,7 +2,8 @@
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "type": "object",
     "properties": {
-        "frame": { "type": "number" }
+        "frame": { "type": "number" },
+        "endframe": {"type": "number"}
```
mwiebe: Daniel and I have discussed how to generalize OpenJD with chunking, and one conclusion we reached is for the chunked version of `onRun` to accept a range expression in a string instead of an integer. We could plan ahead for that development by replacing "frame" with "frameRange", and having it be a range expression conforming to the OpenJD spec. To simplify things, we can require that initially it either is an integer `"<frame>"` or an interval `"<startframe>-<endframe>"`.

If you want to proceed with this, in this PR you would start a two-phase change by making both "frame" and "frameRange" work, while marking "frame" as deprecated. Once that change is rolled out, you would make another PR to remove "frame".

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -121,6 +122,34 @@ def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
             job_template["jobEnvironments"] = []
         job_template["jobEnvironments"].append(override_environment["environment"])
 
+        # Determine whether this is a MOV render. If it is, we want to ensure that the entire Nuke
+        # evaluation is placed on one task.
+        write_node, _ = _get_write_node(settings)
+        mov_render = "file_type" in write_node.knobs() and write_node["file_type"].value() == "mov"
```
mwiebe: Does "mov" fully characterize movie rendering, or is it only for the Quicktime "mov" extension? Are there "mpeg" and other formats too?

mwiebe (changes requested):

src/deadline/nuke_adaptor/NukeAdaptor/schemas/run_data.schema.json
jlangmann-aws: FrameRange likely won't work right now though with parameterization, right, until OpenJD supports chunking?

I could simply replace "endframe" with "frameRange" if it makes the upcoming transition easier. We would only use "frameRange" for "MOV" related renders, and single frame renders would still use "frame". Does that align with what you're thinking?

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
jlangmann-aws: Yes, "mov" characterizes rendering out a file with a "mov" extension. The other available options in Nuke are for rendering single frame images, I don't see any other "movie" related formats (there is no "mpeg" option).

src/deadline/nuke_adaptor/NukeAdaptor/schemas/run_data.schema.json
mwiebe: FrameRange will work fine before OpenJD supports chunking. Replacing both "frame" and "endframe" with "frameRange" will make the upcoming transition smoother. Better to make the one simpler interface instead of bifurcating it.

jlangmann-aws: If we're going to use FrameRange for both mov and single frame workflows, how do we support having this value be two different types? ex: an integer "<frame>" or range string "<startframe-endframe>"? I don't see how we could set a key to have optional types in the run_data.schema.json.

EDIT: ignore the above, looks like multiple types can be defined via `"type": ["number", "string"] ` in the schema.

src/deadline/nuke_adaptor/NukeAdaptor/schemas/run_data.schema.json
```diff
@@ -2,7 +2,7 @@
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "type": "object",
     "properties": {
-        "frame": { "type": "number" }
+        "frameRange": { "type": ["number", "string"] }
```
mwiebe: I think at this point we need to start doing two-phase deployments for changes like this. What that means is the change to do right now is to:

1. Keep "frame", but instead of making it required, make it optional.
2. Add "frameRange", but keep it optional.

This way, the adaptor can accept submissions from the old submitter using "frame" and the new submitter using "frameRange." The change represents the start of phase 1.

Once all the submitters are updated to submit "frameRange" instead of "frame", we will do another PR that removes "frame" from the schema and changes "frameRange" from optional to required.

src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
```diff
@@ -330,7 +333,9 @@ def on_run(self, run_data: dict) -> None:
             raise NukeNotRunningError("Cannot render because Nuke is not running.")
         self.validators.run_data.validate(run_data)
         self._is_rendering = True
-        self._action_queue.enqueue_action(Action("start_render", {"frame": run_data["frame"]}))
+        self._action_queue.enqueue_action(
+            Action("start_render", {"frameRange": run_data["frameRange"]})
```
mwiebe: In the first step of the two-phase deployment, I think something like this would accept both "frameRange" and "frame":

```python
# Before the enqueue_action
if "frameRange" not in run_data and "frame" not in run_data:
    raise ... # exception that one of the two is required.
...
 Action("start_render", {"frameRange": run_data.get("frameRange", str(run_data.get("frame", ""))})
```

src/deadline/nuke_adaptor/NukeClient/nuke_handler.py
```diff
@@ -44,18 +45,38 @@ def __init__(self) -> None:
...
             RuntimeError: If start render is called without a frame number.
         """
-        frame = data.get("frame")
-        if frame is None:
-            raise RuntimeError("NukeClient: start_render called without a frame number.")
+        frame_range = data.get("frameRange")
+        if frame_range is None:
+            raise Exception("NukeClient: start_render called without a frameRange.")
+
+        if isinstance(frame_range, int):
```
mwiebe: Due to the schema enforcement, this will always be a string. Maybe use `if "-" not in frame_range` and then validate that the string can be converted to an integer.

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
mwiebe: I decided to take a peek at that list, and only by going through them one by one discovered that "mxf" is also a movie format. It gives options for Avid codecs.

src/deadline/nuke_submitter/default_nuke_job_template.yaml
```diff
@@ -115,7 +115,7 @@ steps:
     - name: runData
       filename: run-data.yaml
       type: TEXT
-      data: 'frame: {{Task.Param.Frame}}'
+      data: 'frameRange: {{Task.Param.Frame}}'
```
mwiebe: The value here should be a string instead of a number due to the schema. Putting the value in double quotes can do that:

```yaml
 data: 'frameRange: "{{Task.Param.Frame}}"'
```

test/unit/deadline_adaptor_for_nuke/NukeAdaptor/test_adaptor.py
```diff
@@ -43,7 +43,7 @@ def run_data() -> dict:
     Returns:
         dict: A run_data dictionary
     """
-    return {"frame": 42}
+    return {"frameRange": 42}
```
mwiebe: I think this should be `return {"frameRange": "42"}`

mwiebe (changes requested):

src/deadline/nuke_adaptor/NukeAdaptor/schemas/run_data.schema.json
mwiebe: Here's a link to reading material about two-phase deployments if you want to go deeper on the topic: https://aws.amazon.com/builders-library/ensuring-rollback-safety-during-deployments/

src/deadline/nuke_adaptor/NukeClient/nuke_handler.py
jlangmann-aws: I tested this out and the schema enforcement does allow for both int or string. I will go ahead and test to see if the parameterization works with string as well, then we can remove the integer option.

mwiebe (approved):

mwiebe (approved):

amznestebanpapp (approved):

PR #48, opened 2023-10-26
fix: Add default rezPackages + updated deadline dependency to 0.32

jlangmann-aws:
1. RezPackage defaults were removed in a previous PR
2. Deadline-cloud dependency was out of date

1. Added RezPackage defaults `nuke-13 deadline_cloud_for_nuke`
2. Updated deadling-cloud dependency

End user no longer needs to worry about overriding the rez package defaults via the UI

1. Created a new default queue
2. Confirmed that after creating a new scene, rez env is populated correctly.

```
```

### DIFF
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -277,7 +277,10 @@ def show_nuke_render_submitter(parent, f=Qt.WindowFlags()) -> "SubmitJobToDeadli
         g_submitter_dialog = SubmitJobToDeadlineDialog(
             job_setup_widget_type=SceneSettingsWidget,
             initial_job_settings=render_settings,
-            initial_shared_parameter_values={},
+            # Note: nuke-13 is currently the only nuke rez package available. Hardcoding to nuke-13 for now, but
+            # evenutally this should be set to nuke.env["NukeVersionMajor"] once more versions
+            # are supported.
+            initial_shared_parameter_values={"RezPackages": "nuke-13 deadline_cloud_for_nuke"},
             auto_detected_attachments=auto_detected_attachments,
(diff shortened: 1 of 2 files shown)


src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -277,7 +277,10 @@ def on_create_job_bundle_callback(
         g_submitter_dialog = SubmitJobToDeadlineDialog(
             job_setup_widget_type=SceneSettingsWidget,
             initial_job_settings=render_settings,
-            initial_shared_parameter_values={},
+            # Note: nuke-13 is currently the only nuke rez package available. Hardcoding to nuke-13 for now, but
+            # the rez version should eventually be set to nuke.env["NukeVersionMajor"] once more versions
+            # are supported.
+            initial_shared_parameter_values={"RezPackages": "nuke-13 deadline_cloud_for_nuke"},
```
mwiebe: I'd be inclined to use the Nuke major version here already even though we don't have more packages. The user submitting the job can override it in the submitter to nuke-13 if they need to.

mwiebe (approved):

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
jlangmann-aws: Ah ok, yes this is more of a convenience right now since most of us are using nuke-14 locally but the only rez package available is nuke-13. Do we know what version of Nuke beta customers will be running?

jlangmann-aws: I'll update the code with your suggestion, but we should also create a rez package for nuke-14 in the meantime as well.

ttblanchard (approved):

PR #50, opened 2023-10-26
fix: Add optional purpose enum to bundle creation callback.

rmv:

Make Nuke submitter compatible with new purpose argument in job bundle creation callback.

Import the type and add the optional argument.

Make submitter compatible with additional deadline-cloud changes.

Tests pass and it compiles (requires changes in https://github.com/casillas2/deadline-cloud/pull/94

Minor change

No

No


rmv:
NB: this will fail until https://github.com/casillas2/deadline-cloud/pull/94 is in.

agerveshi (approved):

MarinMaksutaj:
Code is already present for this in our main branch. We can resolve.

https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/99dce7eb567644d25e9e9fcf7892bc7293fe4d1f/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py#L379-L386

PR #100, opened 2024-03-04
feat(deps)!: update openjd-adaptor-runtime to 0.5

epmog:
Run is no longer the default adaptor action

Needed to update to the latest openjd adaptor runtime

Updated, and added a SemVer for the openjd interface

Our unversioned API is now versioned

```
hatch run fmt
hatch build
hatch run lint
hatch run test
```

N/A

Yes, nothing to do on the customer's side unless they wrote custom job templates that don't specify an action and expected `run` to happen.

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -8,7 +8,7 @@ sync = "pip install -r requirements-testing.txt"
 test = "pytest {args:test/unit/*}"
 typing = "mypy {args:src test}"
 style = [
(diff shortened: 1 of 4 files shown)


AWS-Samuel (approved):

ddneilson (approved):
Should the title be fix or feat so that it shows up in changelogs?

PR #105, opened 2024-03-13
Include menu.py file in distribution.

rmv:
menu.py was missing from the distribution for deadline-cloud-for-nuke

Added it to pyproject.toml

Users that rely on the pip install will now get the file

created a distribution and the file was there

NO, unrelated change.

No need to.

No.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ include = [
 [tool.hatch.build.targets.wheel]
 packages = [
   "src/deadline",
+  "src/menu.py",
 ]
 
 [tool.mypy]


pyproject.toml
```diff
@@ -56,6 +56,7 @@ include = [
 [tool.hatch.build.targets.wheel]
 packages = [
   "src/deadline",
+  "src/menu.py",
```
epmog, line 59: If we only do this, and perform the pip install, does it live outside the `deadline` namespace package? It should probably be put into the `deadline/nuke_submitter` package. 

If we end up moving the file in source, we'll also have to make sure the installer doesn't break because of it.

amarsjac (approved):

PR #123, opened 2024-03-30
feat!: public release

moorec-aws:

We're about to publicly release this library. We need to prep the build process for that.

- Bump deadline-cloud to `0.47.0`, openjd-adaptor-runtime to `0.7.0` and deadline-test-fixtures to `0.6.0`

- Update the build script to no longer define a PIP_INDEX_URL -- the environment variable
  that we used to interface with our internal repository during private development.

- Update the code quality check to pull deps from the public PyPI, so
  that we're testing as customers would use it.

- Use the public PyPI in the release flows since all deps are now
  available publicly. This ensures that the artifact we release can be
  built & used by anyone using PyPI.

Release readiness

Tests will not pass until dependencies are released

No

No, but it is being marked as one to force a minor version bump when we release.

----


requirements-testing.txt
```diff
@@ -1,3 +1,7 @@
+<<<<<<< HEAD
+=======
+deadline-cloud-test-fixtures == 0.6.*
+>>>>>>> 507ef2c (feat!: public release)
```
jericht: merge conflict

moorec-aws: fixed

AWS-Samuel (approved):

marofke (approved):

amarsjac (approved):

PR #134, opened 2024-04-22
fix: check for video output regardless of adaptor override

epmog:

We had previously added some checks looking for video outputs to not parameterize on frames, but it wasn't working for an internal user.

Turns out the code was accidentally included in the "override adaptor" setting, so it would generally only work for developers.

Just had to de-indent the check to enable it for everyone.

Users can render multi-frame videos with Nuke!

Internal user confirmed that this change made it work for them.

N/A

Nope

----

### DIFF
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -120,34 +120,30 @@ def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
             job_template["jobEnvironments"] = []
         job_template["jobEnvironments"].append(override_environment["environment"])
 
-        # Determine whether this is a movie render. If it is, we want to ensure that the entire Nuke
-        # evaluation is placed on one task.
-        write_node, _ = _get_write_node(settings)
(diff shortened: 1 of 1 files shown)


src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -120,34 +120,30 @@ def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
...
-                if settings.override_frame_range
-                else str(write_node.frameRange())
+    # Determine whether this is a movie render. If it is, we want to ensure that the entire Nuke
+    # evaluation is placed on one task.
+    write_node, _ = _get_write_node(settings)
+    movie_render = "file_type" in write_node.knobs() and write_node["file_type"].value() in [
+        "mov",
+        "mxf",
+    ]
+    if movie_render:
```
ddneilson, line 130: This `_get_template()` function is starting to get long and branchy. Best to nip that in the bud before the pattern of just adding more & more if-else statements takes hold; it should be functionally decomposed in a future change.

ddneilson (approved):

lucaseck (approved):

PR #143, opened 2024-05-14
feat: add ability to configure timeouts in submission UI

AWS-Samuel:

There was no mechanism to set timeouts for session/environment actions through the submitter.

Add the UI. The default looks like the following:
image

* Use Timeouts is a sticky setting
* The values of each timeout type are a sticky setting

A warning pops up when deselecting the `Use Timeouts` checkbox

image

Zero-valued timeouts are not valid, and are highlighted red to help with identification

image

- On submission the following error pops up for zero-valued timeouts
image

The timeout fields are disabled when the `Use Timeouts`  checkbox is not selected

image
- Additionally, the zero-value validation is disabled when this checkbox is deselected

Users can set timeouts in their job and have the corresponding action timeout when the duration has elapsed.

- UI was manually tested
- Submitted a job using these changes and confirmed that the job timed out accordingly

Yes, I updated the job bundle output test job templates as the default value for the timeout fields has changed the expected template 

```

Timestamp: 2024-05-14T11:54:56.189556-05:00
Running job bundle output test: C:\Users\samcan\git\deadline\deadline-cloud-for-nuke\job_bundle_output_tests\cwd-path

cwd-path
Test succeeded

Timestamp: 2024-05-14T11:54:58.856255-05:00
Running job bundle output test: C:\Users\samcan\git\deadline\deadline-cloud-for-nuke\job_bundle_output_tests\group-read

group-read
Test succeeded

Timestamp: 2024-05-14T11:54:59.780206-05:00
Running job bundle output test: C:\Users\samcan\git\deadline\deadline-cloud-for-nuke\job_bundle_output_tests\multi-load-save

multi-load-save
Test succeeded

Timestamp: 2024-05-14T11:55:01.078336-05:00
Running job bundle output test: C:\Users\samcan\git\deadline\deadline-cloud-for-nuke\job_bundle_output_tests\noise-saver

noise-saver
Test succeeded

Timestamp: 2024-05-14T11:55:02.101071-05:00
Running job bundle output test: C:\Users\samcan\git\deadline\deadline-cloud-for-nuke\job_bundle_output_tests\ocio

ocio
Test succeeded

Timestamp: 2024-05-14T11:55:06.716752-05:00

```

Yes

No
----

### DIFF
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -51,6 +51,33 @@ def _get_write_node(settings: RenderSubmitterUISettings) -> tuple[Node, str]:
     return write_node, settings.write_node_selection
 
 
+def _set_timeouts(template: dict[str, Any], settings: RenderSubmitterUISettings) -> None:
+    """
+    Timeouts are an OpenJD field applicable to actions but for specification 2023-09, timeouts must
+    be hard-coded in the job template. There are three types of actions: OnRun, onEnter, and onExit.
+    This function does an in-place modification of timeout values for each action in the template.
+    """
+
+    def _handle_environment(environment: dict):
+        if "script" in environment:
+            actions = environment["script"]["actions"]
+            actions["onEnter"]["timeout"] = settings.on_enter_timeout_seconds
+            if "onExit" in actions:
+                actions["onExit"]["timeout"] = settings.on_exit_timeout_seconds
+
+    def _handle_step(step: dict):
+        for environment in step.get("stepEnvironments", []):
+            _handle_environment(environment)
+
+        step["script"]["actions"]["onRun"]["timeout"] = settings.on_run_timeout_seconds
+
+    for environment in template.get("jobEnvironments", []):
+        _handle_environment(environment)
+
+    for step in template.get("steps", []):
+        _handle_step(step)
+
+
 def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
     # Load the default Nuke job template, and then fill in scene-specific
     # values it needs.
@@ -61,6 +88,9 @@ def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
     job_template["name"] = settings.name
     if settings.description:
         job_template["description"] = settings.description
+
+    # Set the timeouts for each action:
+    _set_timeouts(job_template, settings)
 
     # Get a map of the parameter definitions for easier lookup
     parameter_def_map = {param["name"]: param for param in job_template["parameterDefinitions"]}
@@ -270,6 +300,20 @@ def show_nuke_render_submitter(parent, f=Qt.WindowFlags()) -> "SubmitJobToDeadli
             )
             if result == QMessageBox.Yes:
                 nuke.scriptSave()
+
+        if settings.timeouts_enabled:
+            message = "The following timeout value(s) must be greater than 0: \n"
+            zero_timeouts = []
+            if not settings.on_run_timeout_seconds:
+                zero_timeouts.append("Render Timeout")
+            if not settings.on_enter_timeout_seconds:
+                zero_timeouts.append("Setup Timeout")
+            if not settings.on_exit_timeout_seconds:
+                zero_timeouts.append("Teardown Timeout")
+            if zero_timeouts:
+                message += ", ".join(zero_timeouts)
+                message += "\n\nPlease configure these value(s) in the 'Job-Specific Settings' tab."
+                raise DeadlineOperationError(message)
 
         job_bundle_path = Path(job_bundle_dir)
         job_template = _get_job_template(settings)
--- a/src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
+++ b/src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
@@ -10,11 +10,14 @@ from PySide2.QtWidgets import (  # type: ignore
     QCheckBox,
     QComboBox,
     QGridLayout,
+    QGroupBox,
     QLabel,
     QLineEdit,
+    QMessageBox,
     QSizePolicy,
     QSpacerItem,
     QWidget,
+    QSpinBox,
 )
 
 from ...assets import find_all_write_nodes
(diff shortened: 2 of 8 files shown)


src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -51,6 +51,28 @@ def _get_write_node(settings: RenderSubmitterUISettings) -> tuple[Node, str]:
     return write_node, settings.write_node_selection
 
 
+def _recursive_set_timeouts(
+    data: list[Any] | dict[str, Any], settings: RenderSubmitterUISettings
+) -> None:
+    """
+    Timeouts are an OpenJD field applicable to actions but must be hard-coded in the job template.
```
epmog: maybe put in the spec version or that currently they must be hard-coded? The plan is to propose to change this I believe.

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
```diff
@@ -89,9 +201,27 @@ def refresh_ui(self, settings: RenderSubmitterUISettings):
...
         self.proxy_mode_check.setChecked(settings.is_proxy_mode)
 
+        self.timeout_checkbox.setChecked(settings.timeouts_enabled)
+
+        def _set_timeout(
+            timeout_boxes: tuple[QLabel, QSpinBox, QSpinBox, QSpinBox], timeout_seconds: int
+        ):
+            days = timeout_seconds // 86400
+            hours = (timeout_seconds - days * 86400) // 3600
+            minutes = (timeout_seconds - days * 86400 - hours * 3600) // 60
```
epmog: alternative: you can use the modulo operator here instead of converting the previous days and hours back to seconds each time

```python
# 2 days, 3 hours, 7 minutes
>>> ts = (2 * 86400) + (3 * 3600) + (7 * 60)
>>> ts
184020
>>> (ts % 86400) // 3600
3
>>> (ts % 86400 % 3600 ) // 60
7
```

epmog (dismissed):

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
AWS-Samuel: Added this clarification!

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
AWS-Samuel: Ah, yeah this is way cleaner. Thanks for the suggestion, added!

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -51,6 +51,29 @@ def _get_write_node(settings: RenderSubmitterUISettings) -> tuple[Node, str]:
     return write_node, settings.write_node_selection
 
 
+def _recursive_set_timeouts(
```
crowecawcaw: Why is this function recursive? Do Nuke jobs have multiple render steps?

If so, I think someone might be surprised that after setting a render timeout of 1 day, the job can actually run for several days if there are several render steps.

AWS-Samuel: > Why is this function recursive? 

Since job templates are effectively a tree structure, finding the "actions" within that tree was easiest by recursing through the template. Dynamically finding the actions like this also allows us to modify the job template in the future without being concerned whether we need to change the implementation of this function.

> Do Nuke jobs have multiple render steps?

Right now Nuke jobs only have a single step named `Render` [(link)](https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/src/deadline/nuke_submitter/default_nuke_job_template.yaml#L61)

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
```diff
@@ -48,31 +51,140 @@ def _build_ui(self):
...
+        timeouts_lyt = QGridLayout(self.timeouts_box)
+        lyt.addWidget(self.timeouts_box, 5, 0, 1, -1)
+
+        def create_timeout_row(label, tooltip, row):
+            qlabel = QLabel(label)
+            qlabel.setToolTip(tooltip)
+            timeouts_lyt.addWidget(qlabel, row, 0)
+
+            days_box = QSpinBox(self, minimum=0, maximum=365)
+            days_box.setSuffix(" days")
```
crowecawcaw, line 93: Nit: this text should also support a value of `1`. E.g. "1 day" instead of "1 days"

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
crowecawcaw: Crawling the whole job seems unnecessarily complex. We can just directly access the actions with `job_template["steps"][0]["stepEnvironments"][0]["script"]` right? It also makes unintended side effects less likely since we're specifying the part of the template we want to modify instead of any parameter called "actions".

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
```diff
@@ -89,9 +201,27 @@ def refresh_ui(self, settings: RenderSubmitterUISettings):
...
         self.proxy_mode_check.setChecked(settings.is_proxy_mode)
 
+        self.timeout_checkbox.setChecked(settings.timeouts_enabled)
+
+        def _set_timeout(
+            timeout_boxes: tuple[QLabel, QSpinBox, QSpinBox, QSpinBox], timeout_seconds: int
+        ):
+            days = timeout_seconds // 86400
+            hours = (timeout_seconds % 86400) // 3600
+            minutes = (timeout_seconds % 3600) // 60
```
epmog, line 218: Need to remove the days as well as the hours
```suggestion
            minutes = (timeout_seconds % 86400 % 3600) // 60
```

epmog: Oops nevermind, that still works.

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
epmog: This would apply to hours and minutes as well. Quick change is parenthesis: `day(s)`, better fix is update suffix on value change?

Maybe just the quick change for now?

epmog (dismissed):

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
AWS-Samuel: > better fix is update suffix on value change?

We already had a callback setup for the spin box value change to determine if the sum was 0 valued and mark the label as red. I've added additional logic there to also check if any of the suffixes should be singular

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
AWS-Samuel: >It also makes unintended side effects less likely since we're specifying the part of the template we want to modify instead of any parameter called "actions".

Looking at the current OpenJD spec I don't think that this would be possible because we only look at key names when checking for `action`. Regardless, I didn't want to rule it out completely so I've implemented a non-recursive version of this function which accesses each action directly. It ended up only being a handful of lines longer and is a more explicit which is nice

AWS-Samuel:
diff since last review: https://github.com/aws-deadline/deadline-cloud-for-nuke/compare/baecba26110980f121ced862c971c654f9a81cb7..183e7b96a36e258187c0b60ce4f64043b5af1b44

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
```diff
@@ -48,31 +51,147 @@ def _build_ui(self):
...
+        def hookup_zero_callback(timeout_boxes: tuple[QLabel, QSpinBox, QSpinBox, QSpinBox]):
+            def indicate_is_valid_callback(value: int):
+                self.indicate_if_valid(timeout_boxes)
+
+            for timeout_box in timeout_boxes[1:]:
+                timeout_box.valueChanged.connect(indicate_is_valid_callback)
+
+        self.on_run_timeouts = create_timeout_row(
+            label="Render Timeout",
+            tooltip="Maximum duration of each action which performs a render. Default is 6 days.",
```
crowecawcaw: Is there a way to tweak this text to clarify that it might be per frame? Maybe "Render task timeout" since the user will see the submission represented as a series of tasks in the monitor.

AWS-Samuel: Yeah, I agree - `Render Task Timeout` would be more clear. 

Updated this in the latest commit

crowecawcaw (approved):

epmog (approved):

PR #152, opened 2024-06-27
feat: Added support for including gizmos in job bundle

evanspearman-a:

Nuke gizmos are saved groups of nodes that can be re-used. They are stored in `.gizmo` files. These files must be on the `NUKE_PATH` in order to be used. If a user submits a nuke script that references gizmos and the gizmos are not otherwise installed on the worker, the render will fail when it tries to load the gizmos.

While generating the job bundle, we search the scene for gizmo nodes and collect their file paths. We copy all the utilized gizmos to the job bundle in a `gizmos/` sub directory. We then add a path parameter to the job template referencing this sub directory. We then add a step environment to the Render step _before_ the environment that starts nuke which adds the path mapped gizmos directory to `NUKE_PATH`. There is a checkbox in the submission GUI to disable this if desired.

The reason we copy the gizmo files to the job bundle is that the files are small, and putting them all in one directory rather than uploading each one as individual job attachments and adding each path mapped directory to the `NUKE_PATH` makes the job template more complex.  

Users will be able to submit gizmos with their Nuke jobs.

I submitted a job for a nuke script that used gizmos and verified that it rendered correctly. I also verified that if the option is turned off gizmos are not copied and the step environment and parameter are not added. I also added a unit test for the unit testable code.

Also,  some of the unit tests do not run on Windows, so I have disabled them on Windows.

```log

Timestamp: 2024-06-27T10:48:11.644393-05:00
Running job bundle output test: C:\Users\evans\Development\deadline-cloud-client\deadline-cloud-for-nuke\job_bundle_output_tests\cwd-path

cwd-path
Test succeeded

Timestamp: 2024-06-27T10:48:13.029584-05:00
Running job bundle output test: C:\Users\evans\Development\deadline-cloud-client\deadline-cloud-for-nuke\job_bundle_output_tests\group-read

group-read
Test succeeded

Timestamp: 2024-06-27T10:48:14.994729-05:00
Running job bundle output test: C:\Users\evans\Development\deadline-cloud-client\deadline-cloud-for-nuke\job_bundle_output_tests\multi-load-save

multi-load-save
Test succeeded

Timestamp: 2024-06-27T10:48:16.213881-05:00
Running job bundle output test: C:\Users\evans\Development\deadline-cloud-client\deadline-cloud-for-nuke\job_bundle_output_tests\noise-saver

noise-saver
Test succeeded

Timestamp: 2024-06-27T10:48:18.216164-05:00
Running job bundle output test: C:\Users\evans\Development\deadline-cloud-client\deadline-cloud-for-nuke\job_bundle_output_tests\ocio

ocio
Test succeeded

Timestamp: 2024-06-27T10:48:25.738289-05:00
```

No

No

----

### DIFF
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import os
 import re
+import shutil
 from pathlib import Path
 from typing import Any, Optional
 import yaml  # type: ignore[import]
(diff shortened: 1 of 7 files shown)


AWS-Samuel (approved):

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -78,7 +79,70 @@ def _handle_step(step: dict):
...
+    render_step = None
+    for step in job_template["steps"]:
+        if step["name"] == "Render":
+            render_step = step
+            break
+    if render_step is None:
+        raise RuntimeError("No 'Render' step found in the job template")
+
+    if "stepEnvironments" not in render_step:
+        render_step["stepEnvironments"] = []
```
epmog, line 118: I wonder if this all just makes more sense as a jobEnvironment for the jobs created by the nuke submitter. I think we only have 1 step right now but that might change in the future with the hard-coded step name also then changing...

I guess keeping it as a step environment allows users to export the bundle and add their own step without being forced to add the gizmos to their non-nuke step.

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -78,7 +79,70 @@ def _handle_step(step: dict):
...
+
+    # This needs to be prepended rather than appended
+    # as it must run before the "Nuke" environment.
+    render_step["stepEnvironments"].insert(
+        0,
+        {
+            "name": "AddGizmoPathsToNukePath",
```
epmog, line 125: nit: this will be read by users, not necessarily developers. So we should add spaces here and a description that will show up in the service:

```
Add Gizmos to NUKE_PATH
```
or something like that

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -78,7 +79,70 @@ def _handle_step(step: dict):
...
+def _add_gizmo_dir_to_job_template(job_template: dict[str, Any], relative_gizmo_dir: Path) -> None:
+    if "parameterDefinitions" not in job_template:
+        job_template["parameterDefinitions"] = []
+
+    job_template["parameterDefinitions"].append(
+        {
+            "name": "GizmoDir",
+            "type": "PATH",
+            "description": "The directory containing Nuke Gizmo files used by this job.",
+            "default": str(relative_gizmo_dir),
```
epmog, line 102: Any reason to not hard-code the default to `./gizmos` and instead add the gizmopath to the settings which would then allow a user to provide a different value via the parameter_values to the job bundle? Would prevent the need to modify and pass this directory around that's always going to be the same for this code

epmog (commented):
Thanks for the PR Evan! It looks like the feature works in the job bundle, my comments are more about how we structure the resulting job bundle to allow easier manual tweaking.

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
diane-hj: Thanks for the feedback. I addressed these comments in this PR: https://github.com/aws-deadline/deadline-cloud-for-nuke/pull/162

epmog:
The other PR that took over this work as been merged. Thanks Evan! Closing.

PR #154, opened 2024-07-10
fix: update frame range when write node is selected

diane-hj:
- When write node is selected, incorrect frame range were exported in parameters

- When override frame is selected, override the frame
- When write node is selected with limit_range checked (without override), write node's frame applied
- When write node is selected without limit_range checked (without override), project settings frame applied
- When no specific write node is selected, project settings frame applied

- Based on the customers' setting, right frame range is selected as explained above 

- Opened Nuke with the following setting: 
```
Project setting frame: 1-4
Write1 node frame: 1-10
Override when selected with frame: 4
```
Verified the following: 
```
All write nodes, without frame override —> 1 - 4 
All write nodes, with frame override —> 4 
All write nodes, without frame override —> 1 - 4

Write1, with limitRange, without frame override —> 1 - 10 
Write1, with limitRange, with frame override —> 4
Write1, with limitRange, without frame override —> 1- 10 

Write1, without limitRange, without frame override —> 1- 4
Write1, without limitRange, with frame override —> 4 
Write1, without limitRange, without frame override —> 1 - 4 

```

```

Timestamp: 2024-08-12T14:01:24.012286-07:00

group-read
Test succeeded

Timestamp: 2024-08-12T14:01:25.055533-07:00

noise-saver
Test succeeded

Timestamp: 2024-08-12T14:01:26.720246-07:00

multi-load-save
Test succeeded

Timestamp: 2024-08-12T14:01:27.757395-07:00

frame
Test succeeded

Timestamp: 2024-08-12T14:01:29.360422-07:00

ocio
Test succeeded

Timestamp: 2024-08-12T14:01:30.340508-07:00

cwd-path
Test succeeded

Timestamp: 2024-08-12T14:01:33.469925-07:00

```

I don't think so. 
This is a fix. 

----

### DIFF
--- a/job_bundle_output_tests/frame/expected_job_bundle/asset_references.yaml
+++ b/job_bundle_output_tests/frame/expected_job_bundle/asset_references.yaml
@@ -1,0 +1,9 @@
+assetReferences:
+  inputs:
+    directories: []
+    filenames:
+    - /normalized/job/bundle/dir/frame.nk
+  outputs:
+    directories:
+    - /normalized/cwd
+  referencedPaths: []
--- a/job_bundle_output_tests/frame/expected_job_bundle/parameter_values.yaml
+++ b/job_bundle_output_tests/frame/expected_job_bundle/parameter_values.yaml
@@ -1,0 +1,17 @@
+parameterValues:
+- name: Frames
+  value: 3-7
+- name: NukeScriptFile
+  value: /normalized/job/bundle/dir/frame.nk
+- name: WriteNode
+  value: Write1
+- name: ProxyMode
+  value: 'false'
+- name: deadline:targetTaskRunStatus
+  value: READY
+- name: deadline:maxFailedTasksCount
+  value: 20
+- name: deadline:maxRetriesPerTask
+  value: 5
+- name: deadline:priority
+  value: 50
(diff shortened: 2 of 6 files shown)


epmog (changes requested):
Thanks for the PR!

I believe this is the type of change we need since the write node frame range is bugged, but I think there's an additional section we have to update here that will have the exact same problem: https://github.com/aws-deadline/deadline-cloud-for-nuke/pull/154/files#diff-fb4f3580c4b9cd3e4360d8e095ffc997ebf2a140868124094ddbb8755d3ee51fR155-R157

That section there is ran when the user is specifying a write node (rather than all of them).

epmog:
It probably makes sense to include this as one of the job bundle tests as well to help ensure we don't regress.

The job bundle tests are basically structured in such a way that given a scene file, we can verify the output bundle that gets created. So if we have a write node that has different frames from the project, we are then testing it with the expected bundle matching the write node.

Let me know your thoughts

epmog (approved):
Thanks for making those changes. LGTM!

Cherie-Chen (approved):
Looks great! Thank you for adding a through test for this fix!

PR #160, opened 2024-08-20
revert: "fix: update frame range when write node is selected (#154)"

evanspearman-a:
Reverting a PR that was mistakenly merged during the release window.


diane-hj (approved):
Sorry! Thanks for reverting the PR.

lucaseck (approved):

PR #162, opened 2024-08-29
feat: Added support for including gizmos in job bundle

diane-hj:

This PR originated from @evanspearman-a 's PR: https://github.com/aws-deadline/deadline-cloud-for-nuke/pull/152
Most of updates are same from the above PR and credits go to @evanspearman-a.  This PR is just to address comments from the above PR. 

- Updated default_nuke_job_template.yaml to add Gizmo directory
- Moved Gizmo environment from step to job
- Updated human readable step environments

- When Gizmo is enabled, default gizmo path will be in template and gizmo environment will be defined in job environment

- When ExportToBundle with gizmo enabled, verified the template has the gizmo information 
- Submitted colorbars.nk and verified that it has `Launch Add Gizmos to NUKE_PATH` session before launching Nuke 
- Verified that when gizmo is not enabled, those information is not included in template

Screenshot 2024-08-29 at 3 30 39 PM

```

Timestamp: 2024-08-29T15:22:16.581585-07:00

group-read
Test succeeded

Timestamp: 2024-08-29T15:22:17.641190-07:00

noise-saver
Test succeeded

Timestamp: 2024-08-29T15:22:17.965391-07:00

multi-load-save
Test succeeded

Timestamp: 2024-08-29T15:22:18.309697-07:00

frame
Test succeeded

Timestamp: 2024-08-29T15:22:18.651177-07:00

ocio
Test succeeded

Timestamp: 2024-08-29T15:22:18.995647-07:00

cwd-path
Test succeeded

Timestamp: 2024-08-29T15:22:22.044715-07:00

```

No 

No 

----

### DIFF
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -78,6 +78,40 @@ def _set_timeouts(template: dict[str, Any], settings: RenderSubmitterUISettings)
         _handle_step(step)
 
 
+def _remove_gizmo_dir_from_job_template(job_template: dict[str, Any]) -> None:
+    for index, param in enumerate(job_template["parameterDefinitions"]):
+        if param["name"] == "GizmoDir":
+            job_template["parameterDefinitions"].pop(index)
+            break
+
+
+def _add_gizmo_dir_to_job_template(job_template: dict[str, Any]) -> None:
+    if "jobEnvironments" not in job_template:
+        job_template["jobEnvironments"] = []
+
+    # This needs to be prepended rather than appended
+    # as it must run before the "Nuke" environment.
+    job_template["jobEnvironments"].insert(
+        0,
+        {
+            "name": "Add Gizmos to NUKE_PATH",
+            "script": {
+                "actions": {"onEnter": {"command": "{{Env.File.Enter}}"}},
+                "embeddedFiles": [
+                    {
+                        "name": "Enter",
+                        "type": "TEXT",
+                        "runnable": True,
+                        "data": """#!/bin/bash
+    echo 'openjd_env: NUKE_PATH=$NUKE_PATH:{{Param.GizmoDir}}'
+    """,
+                    }
+                ],
+            },
+        },
+    )
+
+
 def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
     # Load the default Nuke job template, and then fill in scene-specific
     # values it needs.
@@ -91,6 +125,13 @@ def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
 
     # Set the timeouts for each action:
     _set_timeouts(job_template, settings)
+
+    # Add Gizmo directory to NUKE_PATH if we copied
+    # any gizmos to the job bundle.
+    if settings.include_gizmos_in_job_bundle:
+        _add_gizmo_dir_to_job_template(job_template)
+    else:
+        _remove_gizmo_dir_from_job_template(job_template)
 
     # Get a map of the parameter definitions for easier lookup
     parameter_def_map = {param["name"]: param for param in job_template["parameterDefinitions"]}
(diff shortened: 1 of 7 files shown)


src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -78,6 +78,40 @@ def _handle_step(step: dict):
         _handle_step(step)
 
 
+def _remove_gizmo_dir_from_job_template(job_template: dict[str, Any]) -> None:
```
crowecawcaw, line 81: Could you also add a unit test for this function?

diane-hj: Yep. Added the unit test for removing gizmo from job template. Thanks!

crowecawcaw (dismissed):

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
```diff
@@ -76,6 +76,9 @@ def _build_ui(self):
         timeouts_lyt = QGridLayout(self.timeouts_box)
         lyt.addWidget(self.timeouts_box, 5, 0, 1, -1)
 
+        self.gizmos_checkbox = QCheckBox("Include Gizmos In Job Bundle", self)
```
epmog, line 79: Hmm... The labels for all of this _should_ be sentence-cased, but I see others that aren't. Let's not deal with that here though.

test/unit/nuke_util/test_ocio.py
```diff
@@ -109,6 +109,7 @@ def test_create_config_from_file(
     create_from_file.assert_called_once_with(custom_ocio_config_path)
 
 
+@pytest.mark.skipif(os.name == "nt", reason="Does not run on Windows")
```
epmog: I see this is from the previous PR, but I'm wondering what error we're seeing because I'd like to avoid commenting out a test on an OS without understanding why.

My hunch is that the mock for the OCIO config isn't set-up correctly for windows:

https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/35b9ceb28598dacfa7337e717681f91e01e57228/test/unit/nuke_util/test_ocio.py#L14-L18

and that we're failing on this check because of that:

https://github.com/diane-hj/deadline-cloud-for-nuke/blob/35b9ceb28598dacfa7337e717681f91e01e57228/src/deadline/nuke_util/ocio.py#L30-L32

`PurePath` will evaluate to the current OS flavour, these absolute posix paths don't look absolute on windows. Here's a Local run from my machine:
```
>>> PurePath("/asdf/asdf")
PurePosixPath('/asdf/asdf')
>>> PurePath("/asdf/asdf").is_absolute()
True
>>> PureWindowsPath("/asdf/asdf").is_absolute()
False
```

So the only thing we should need to do here to get this test to run is to conditionally set the mock for it's OS. Example:

```python
if os.name == "nt":
    return MockOCIOConfig(
        working_dir="C:/this/ocio_configs", search_paths=["luts", "C:/my/absolute/luts"]
    )
return MockOCIOConfig(
        working_dir="/this/ocio_configs", search_paths=["luts", "/my/absolute/luts"]
    )
```

epmog (changes requested):
Thanks for taking on those changes! I have one comment about a change from the previous author that I believe we can easily change to get more tests running on Windows.

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
epmog: Created https://github.com/aws-deadline/deadline-cloud-for-nuke/issues/163. Resolving

test/unit/nuke_util/test_ocio.py
diane-hj: Thanks Morgan! As discussed offline, I created a separate Github issue for this: https://github.com/aws-deadline/deadline-cloud-for-nuke/issues/164

crowecawcaw (approved):

epmog (approved):

PR #166, opened 2024-09-17
fix: Add OCIO configuration to Job Environments

diane-hj:
- There are 3 ways to set OCIO configuration in Nuke: environment variable, custom config, and default config that is defined in Nuke already. The original implementation only took care of custom config. We want to handle when customer set their OCIO configuration in environment variable or default config. 

- When OCIO is enabled, check the configuration path for all three options, then set the path via JobEnvironment in JobTemplate. 
 
- When environment variable is set, the path will be set with environment variable 
- When custom config is set, the path will be set with custom config
- When default config is set, the path will be set with default config 
 
- Submitted the job with OCIO setup, and verified it rendered successfully 
- Submitted the above job with different OCIO setup, verified it rendered successfully, and verified the output looks different in color scheme 
- Submitted the job with OCIO environment variable, and verified it rendered successfully 

```

Timestamp: 2024-09-16T22:51:08.146094-07:00

group-read
Test succeeded

Timestamp: 2024-09-16T22:51:08.578900-07:00

noise-saver
Test succeeded

Timestamp: 2024-09-16T22:51:08.968215-07:00

multi-load-save
Test succeeded

Timestamp: 2024-09-16T22:51:09.358968-07:00

frame
Test succeeded

Timestamp: 2024-09-16T22:51:09.757446-07:00

ocio
Test succeeded

Timestamp: 2024-09-16T22:51:10.157282-07:00

cwd-path
Test succeeded

Timestamp: 2024-09-16T22:51:13.578587-07:00

```

No 

No - In Nuke, environment variable takes precedence over other ocio configurations. (Ex. When OCIO environment variable is set, updating custom config in UI is disabled with the message: `OCIO environment variable is set. Setting OCIO configs from within Nuke will be disabled.`)

The existing user who uses custom config will continue to work. 

----

### DIFF
--- a/job_bundle_output_tests/ocio/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/ocio/expected_job_bundle/template.yaml
@@ -16,6 +16,11 @@ parameterDefinitions:
       patterns:
       - '*'
   description: The Nuke script file to render.
+- name: OCIOConfigPath
+  type: PATH
+  objectType: FILE
+  dataFlow: IN
+  description: The directory containing OCIO config files used by this job.
 - name: Frames
   type: STRING
   description: The frames to render. E.g. 1-3,8,11-15
@@ -126,3 +131,15 @@ steps:
         cancelation:
           mode: NOTIFY_THEN_TERMINATE
         timeout: 518400
+jobEnvironments:
+- name: Add OCIO Path to Environment Variable
+  script:
+    actions:
+      onEnter:
+        command: '{{Env.File.Enter}}'
+    embeddedFiles:
+    - name: Enter
+      type: TEXT
+      runnable: true
+      data: "#!/bin/bash\n    echo 'openjd_env: OCIO={{Param.OCIOConfigPath}}'\n \
+        \   "
--- a/job_bundle_output_tests/ocio/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/ocio/expected_job_bundle/template.yaml
@@ -20,7 +20,7 @@ parameterDefinitions:
   type: PATH
   objectType: FILE
   dataFlow: IN
-  description: The directory containing OCIO config files used by this job.
+  description: The OCIO config file used by this job.
 - name: Frames
   type: STRING
   description: The frames to render. E.g. 1-3,8,11-15
@@ -133,13 +133,5 @@ steps:
         timeout: 518400
 jobEnvironments:
 - name: Add OCIO Path to Environment Variable
-  script:
-    actions:
-      onEnter:
-        command: '{{Env.File.Enter}}'
-    embeddedFiles:
-    - name: Enter
-      type: TEXT
-      runnable: true
-      data: "#!/bin/bash\n    echo 'openjd_env: OCIO={{Param.OCIOConfigPath}}'\n \
-        \   "
+  variables:
+    OCIO: '{{Param.OCIOConfigPath}}'
--- a/src/deadline/nuke_submitter/assets.py
+++ b/src/deadline/nuke_submitter/assets.py
@@ -7,7 +7,7 @@ import re
 from collections.abc import Generator
 from os.path import commonpath, dirname, join, normpath, samefile
 from sys import platform
-
+from typing import Optional
 import nuke
 
 from deadline.client.job_bundle.submission import AssetReferences
@@ -81,17 +81,41 @@ def get_scene_asset_references() -> AssetReferences:
             for filename in get_node_filenames(node):
                 asset_references.output_directories.add(dirname(filename))
 
-    # if using a custom OCIO config, add the config file and associated search directories
-    if nuke_ocio.is_custom_config_enabled():
-        ocio_config_path = nuke_ocio.get_custom_config_path()
-        ocio_config_search_paths = nuke_ocio.get_config_absolute_search_paths(ocio_config_path)
+    if nuke_ocio.is_OCIO_enabled():
+        # Determine and add the config file and associated search directories
+        ocio_config_path = get_ocio_config_path()
+        # Add the references
+        if ocio_config_path is not None:
+            if os.path.isfile(ocio_config_path):
+                asset_references.input_filenames.add(ocio_config_path)
 
-        asset_references.input_filenames.add(ocio_config_path)
-
-        for search_path in ocio_config_search_paths:
-            asset_references.input_directories.add(search_path)
+                ocio_config_search_paths = nuke_ocio.get_config_absolute_search_paths(
+                    ocio_config_path
+                )
+                for search_path in ocio_config_search_paths:
+                    asset_references.input_directories.add(search_path)
+            else:
+                nuke.alert(
+                    "OCIO config file specified(%s) is not an existing file." % ocio_config_path
+                )
 
     return asset_references
+
+
+def get_ocio_config_path() -> Optional[str]:
+    ocio_config_path = None
+
+    # if using a custom OCIO environment variable
+    if nuke_ocio.is_env_config_enabled():
+        ocio_config_path = nuke_ocio.get_env_config_path()
+    else:
+        print("here?")
+        # if using a custom OCIO config file
+        if nuke_ocio.is_custom_config_enabled():
+            ocio_config_path = nuke_ocio.get_custom_config_path()
+        elif nuke_ocio.is_stock_config_enabled():
+            ocio_config_path = nuke_ocio.get_stock_config_path()
+    return ocio_config_path
 
 
 def find_all_write_nodes() -> set:
--- a/src/deadline/nuke_submitter/assets.py
+++ b/src/deadline/nuke_submitter/assets.py
@@ -95,27 +95,24 @@ def get_scene_asset_references() -> AssetReferences:
                 for search_path in ocio_config_search_paths:
                     asset_references.input_directories.add(search_path)
             else:
-                nuke.alert(
-                    "OCIO config file specified(%s) is not an existing file." % ocio_config_path
+                raise DeadlineOperationError(
+                    "OCIO config file specified(%s) is not an existing file. Please check and update the config file before proceeding."
+                    % ocio_config_path
                 )
 
     return asset_references
 
 
 def get_ocio_config_path() -> Optional[str]:
-    ocio_config_path = None
-
     # if using a custom OCIO environment variable
     if nuke_ocio.is_env_config_enabled():
-        ocio_config_path = nuke_ocio.get_env_config_path()
+        return nuke_ocio.get_env_config_path()
+    elif nuke_ocio.is_custom_config_enabled():
+        return nuke_ocio.get_custom_config_path()
+    elif nuke_ocio.is_stock_config_enabled():
+        return nuke_ocio.get_stock_config_path()
     else:
-        print("here?")
-        # if using a custom OCIO config file
-        if nuke_ocio.is_custom_config_enabled():
-            ocio_config_path = nuke_ocio.get_custom_config_path()
-        elif nuke_ocio.is_stock_config_enabled():
-            ocio_config_path = nuke_ocio.get_stock_config_path()
-    return ocio_config_path
+        return None
 
 
 def find_all_write_nodes() -> set:
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -16,6 +16,7 @@ from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import (  # type: 
     SubmitJobToDeadlineDialog,
     JobBundlePurpose,
 )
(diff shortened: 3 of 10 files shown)


job_bundle_output_tests/ocio/expected_job_bundle/template.yaml
```diff
@@ -126,3 +131,15 @@ steps:
...
+- name: Add OCIO Path to Environment Variable
+  script:
+    actions:
+      onEnter:
+        command: '{{Env.File.Enter}}'
+    embeddedFiles:
+    - name: Enter
+      type: TEXT
+      runnable: true
+      data: "#!/bin/bash\n    echo 'openjd_env: OCIO={{Param.OCIOConfigPath}}'\n \
```
eherozhao: nit: I think we don't need to have `\n`
It can be 
```
...
    runnable: true
    data: | 
        #!/bin/bash
        echo 'openjd_env: OCIO={{Param.OCIOConfigPath}}'
```

src/deadline/nuke_submitter/assets.py
```diff
@@ -81,17 +81,41 @@ def get_scene_asset_references() -> AssetReferences:
...
 
-    # if using a custom OCIO config, add the config file and associated search directories
-    if nuke_ocio.is_custom_config_enabled():
-        ocio_config_path = nuke_ocio.get_custom_config_path()
-        ocio_config_search_paths = nuke_ocio.get_config_absolute_search_paths(ocio_config_path)
+    if nuke_ocio.is_OCIO_enabled():
+        # Determine and add the config file and associated search directories
+        ocio_config_path = get_ocio_config_path()
+        # Add the references
+        if ocio_config_path is not None:
```
crowecawcaw, line 88: Is the ocio config path relative and not absolute? Are there any issues with path mapping or anything like that?

src/deadline/nuke_submitter/assets.py
```diff
@@ -81,17 +81,41 @@ def get_scene_asset_references() -> AssetReferences:
...
 
-    return asset_references
+def get_ocio_config_path() -> Optional[str]:
+    ocio_config_path = None
+
+    # if using a custom OCIO environment variable
+    if nuke_ocio.is_env_config_enabled():
+        ocio_config_path = nuke_ocio.get_env_config_path()
+    else:
+        print("here?")
```
crowecawcaw: Nit: remove

src/deadline/nuke_submitter/assets.py
```diff
@@ -81,17 +81,41 @@ def get_scene_asset_references() -> AssetReferences:
...
-        for search_path in ocio_config_search_paths:
-            asset_references.input_directories.add(search_path)
 
-    return asset_references
+def get_ocio_config_path() -> Optional[str]:
+    ocio_config_path = None
+
+    # if using a custom OCIO environment variable
+    if nuke_ocio.is_env_config_enabled():
+        ocio_config_path = nuke_ocio.get_env_config_path()
```
crowecawcaw: Nit: maybe cleaner if we just return the first discovered OCIO config instead of nesting ifs

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -239,6 +285,12 @@ def _get_parameter_values(
     parameter_values.append(
         {"name": "ProxyMode", "value": "true" if settings.is_proxy_mode else "false"}
     )
+
+    # Set the OCIO config path value
+    if nuke_ocio.is_OCIO_enabled():
+        ocio_config_path = get_ocio_config_path()
+        parameter_values.append({"name": "OCIOConfigPath", "value": ocio_config_path})
```
crowecawcaw: Should we check if the path is `None`?

src/deadline/nuke_submitter/assets.py
```diff
@@ -81,17 +81,41 @@ def get_scene_asset_references() -> AssetReferences:
...
+
+                ocio_config_search_paths = nuke_ocio.get_config_absolute_search_paths(
+                    ocio_config_path
+                )
+                for search_path in ocio_config_search_paths:
+                    asset_references.input_directories.add(search_path)
+            else:
+                nuke.alert(
+                    "OCIO config file specified(%s) is not an existing file." % ocio_config_path
+                )
```
crowecawcaw: Should this error block the submission? What happens if it's missing?

src/deadline/nuke_submitter/assets.py
diane-hj: ocio config path is in absolute. `get_ocio_config_path()` will return absolute path for stock and custom configurations. Yeah, when testing, it didn't work properly with relative path. :) Please let me know if you have any concerns.

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -112,6 +118,40 @@ def _add_gizmo_dir_to_job_template(job_template: dict[str, Any]) -> None:
...
+                    {
+                        "name": "Enter",
+                        "type": "TEXT",
+                        "runnable": True,
+                        "data": """#!/bin/bash
+    echo 'openjd_env: OCIO={{Param.OCIOConfigPath}}'
+    """,
+                    }
+                ],
+            },
```
crowecawcaw: I think we can define environment variables directly in OJD instead of with a script: https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#4-environment

I haven't used it personally, but it looks like it'd do what we want.

src/deadline/nuke_util/ocio.py
```diff
@@ -4,11 +4,40 @@
...
+    return "OCIO" in os.environ
+
+
+def get_env_config_path() -> Optional[str]:
+    """This is the path to the custom OCIO config used by the OCIO env var."""
+    return os.environ.get("OCIO")
+
+
+def is_stock_config_enabled() -> bool:
+    """True if the script is using a custom OCIO config"""
```
crowecawcaw: Comment needs to be updated

crowecawcaw (changes requested):

src/deadline/nuke_submitter/assets.py
diane-hj: Dang! I thought I erased all of it. haha Thanks for catching it.

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
diane-hj: Nice! Thanks, its much simpler. 🙏

src/deadline/nuke_submitter/assets.py
diane-hj: If its missing, it means that OCIO is enabled but the provided file is not an existing file or working file. In this case, it will not open Submitter pop up and instead will show the alert message. 
Screenshot 2024-09-17 at 11 29 59 AM

diane-hj: Actually, I take this back. It will still show the Submitter pop up after this alert. I will make changes to not show the submitter if the OCIO config file is not specified properly.

job_bundle_output_tests/ocio/expected_job_bundle/template.yaml
diane-hj: This block is deleted :).

src/deadline/nuke_submitter/assets.py
diane-hj: Updated. Please let me know if I misunderstood your comment 🙏

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
diane-hj: Although it should not happen when OCIO is enabled, I added safe guard for this just in case, somehow it happens, customers would know what they should do to unblock themselves. Thanks!

crowecawcaw (dismissed):

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -144,6 +172,12 @@ def _get_job_template(settings: RenderSubmitterUISettings) -> dict[str, Any]:
     # Set the View parameter allowed values
     parameter_def_map["View"]["allowedValues"] = ["All Views"] + sorted(nuke.views())
 
+    # if OCIO is disabled, remove OCIO path from the template
+    if nuke_ocio.is_OCIO_enabled():
+        _add_ocio_path_to_job_template(job_template)
+    else:
+        _remove_ocio_path_from_job_template(job_template)
```
lucaseck, line 179: What's the use case for removing it? Is the OCIO path already there in the template, I thought it was only added if OCIO was enabled two lines above. Can the param only be added if OCIO is being used and the default template left the same?

src/deadline/nuke_submitter/default_nuke_job_template.yaml
```diff
@@ -25,6 +25,11 @@ parameterDefinitions:
     label: Gizmo Directory
   description: The directory containing Nuke Gizmo files used by this job.
   default: './gizmos'
+- name: OCIOConfigPath
+  type: PATH
+  objectType: FILE
+  dataFlow: IN
+  description: The directory containing OCIO config files used by this job.
```
lucaseck: This looks to be just a single FILE not a directory like the description says. One or the other probably needs to be updated.

job_bundle_output_tests/ocio/expected_job_bundle/template.yaml
lucaseck: I might be missing some context, this block still looks like it exists and has the `\n` everywhere. Is that just a consequence of the environment being written out or do we have control over what this block looks like in the generated template? If we have control then personally I prefer the way that Hao outlined.

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
diane-hj: The remove method is to remove any OCIO information from the job_template's parameter definitions. If there is, I think it will complain about not having the default value, etc. Also, I thought if its not being used, it is ideal to remove it from the parameter definition. Please let me know how you think about it 🙏 

The upper part of adding is to add the `jobEnvironment` definition for it using the parameter definition that is in default job template. Please let me know if it makes sense!

src/deadline/nuke_submitter/default_nuke_job_template.yaml
diane-hj: Thanks for catching, I will update this to something like, 
`description: The OCIO config file used by this job`

job_bundle_output_tests/ocio/expected_job_bundle/template.yaml
diane-hj: Oh no, this part was updated but not included in the commit. I will add it in the next commit. This was updated to something like 
```
jobEnvironments:
- name: Add OCIO Path to Environment Variable
  variables:
    OCIO: '{{Param.OCIOConfigPath}}'
```

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
lucaseck: Can we remove the parameter entirely from the default job template though? Only add the parameter into the template if OCIO is enabled. Then the remove method wouldn't be needed at all and OCIO relevant stuff would only be included if enabled

lucaseck: Spoke offline, personally I'm of the opinion that the parameter without the environment and vice versa is not really useful so I don't see why we would bake one into the default template, but not the other. Given Gizmo handling already follows this pattern though I'm not going to block the review on it, but leave it up to the owners to decide whether both OCIO and Gizmo parameters should be in the default template or not.

src/deadline/nuke_submitter/default_nuke_job_template.yaml
```diff
@@ -25,6 +25,11 @@ parameterDefinitions:
     label: Gizmo Directory
   description: The directory containing Nuke Gizmo files used by this job.
   default: './gizmos'
+- name: OCIOConfigPath
+  type: PATH
+  objectType: FILE
+  dataFlow: IN
+  description: The OCIO config file used by this job.
```
lucaseck, line 32: Is there a default that can make sense here?

diane-hj: Someone else can correct me if I'm wrong. 🙏 
I don't think we have a perfect default path T-T because it's really up to customers. Customers can bring their configuration by themselves or use Nuke's default configurations where Nuke's default has more than 3 configuration that customers can choose so it is hard to choose one.

lucaseck: Makes sense 👍

lucaseck (approved):

crowecawcaw (approved):

PR #199, opened 2025-03-11
test: ensure schema changes require version bump

### LINKED ISSUE #191: Maintenance: add test for `integration_data_interface_version`
### Description

When the `init_data.schema.yaml` or `run_data.schema.yaml` is changed, the adaptor's `integration_data_interface_version` should also be increased. We have a note in the pull request template that this is a breaking change, but the instructions only indicate to include an `!` in your conventional commit title and to add instructions for upgrading to the PR. However, they don't indicate that you need to increase the `integration_data_interface_version`, which could be missed.

We should add tests that if the `init_data.schema.yaml` or `run_data.schema.yaml` is changed, the `integration_data_interface_version` is also increased.

### Solution

[Here is the test we added for a different Deadline Cloud integration for this](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/51146c450a96994ac016083740f289edc1e83824/test/unit/keyshot_adaptor/KeyShotAdaptor/test_adaptor.py#L56). A similar test can be added to this repository.

MarinMaksutaj:
Fixes #191

This is a code change for Issue #191. When schema files are modified, there was no automated way to ensure that the integration_data_interface_version is also bumped to maintain backward compatibility.

Added a test that compares the current schema files with stored versions and verifies that if they change, the integration_data_interface_version is also bumped accordingly.

This change helps prevent unintentional breaking changes by ensuring schema modifications are properly versioned, maintaining backward compatibility for clients.

The test was run both in isolation and as part of the full test suite to verify it correctly detects schema changes and version requirements.

No.

-

No, this is a non-breaking change that adds test coverage.

### DIFF
--- a/test/unit/deadline_adaptor_for_nuke/NukeAdaptor/test_schema_versioning.py
+++ b/test/unit/deadline_adaptor_for_nuke/NukeAdaptor/test_schema_versioning.py
@@ -1,0 +1,85 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+
+from deadline.nuke_adaptor.NukeAdaptor.adaptor import NukeAdaptor
+
+
+@pytest.fixture
+def init_data():
+    """Fixture providing basic init data for the adaptor"""
+    return {"script_file": "test.nk"}
+
+
+def test_schema_matches_reference_version(init_data):
+    """
+    Test to validate that the schema matches the reference version.
+    When schemas are modified, both the reference schemas AND the expected version
+    should be updated together in this test.
+    """
+    # Define reference schemas that represent the current expected state
+    REFERENCE_INIT_DATA_SCHEMA = {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "type": "object",
+        "properties": {
+            "continue_on_error": {"type": "boolean"},
+            "proxy": {"type": "boolean"},
+            "write_nodes": {"type": "array", "items": {"type": "string"}},
+            "views": {"type": "array", "items": {"type": "string"}},
+            "telemetry_opt_out": {"type": "boolean"},
+            "script_file": {"type": "string"},
+        },
+        "required": ["script_file"],
+    }
+
+    REFERENCE_RUN_DATA_SCHEMA = {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "type": "object",
+        "properties": {"frame": {"type": "number"}, "frameRange": {"type": "string"}},
+        "optional": ["frameRange", "frame"],
+    }
+
+    # Expected version for these reference schemas
+    EXPECTED_MAJOR = 0
+    EXPECTED_MINOR = 1
+
+    # Get the current version from the adaptor
+    adaptor = NukeAdaptor(init_data)
+    semantic_version = adaptor.integration_data_interface_version
+
+    # Load current schemas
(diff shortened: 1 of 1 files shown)


test/unit/deadline_adaptor_for_nuke/NukeAdaptor/test_schema_versioning.py
```diff
@@ -0,0 +1,82 @@
...
+
+def test_if_init_data_and_run_data_schema_are_changed_schema_version_is_bumped(init_data):
+    """
+    Test to validate that if the init data or run data schema are changed, we also bump the
+    integration_data_interface_version
+    """
+    adaptor = NukeAdaptor(init_data)
+    semantic_version = adaptor.integration_data_interface_version
+
+    root_directory_path = Path(__file__).parent.parent.parent.parent.parent
```
viknith, line 56: Won't this test always pass since we're just loading the same JSONs twice and verifying that they're equal? It looks like `CURRENT_INIT_DATA_SCHEMA` and `init_data_schema` from the same place so they'll end up equal.

MarinMaksutaj: That's true.. thank you for the catch, will make a revision in a bit.

test/unit/deadline_adaptor_for_nuke/NukeAdaptor/test_schema_versioning.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def test_if_init_data_and_run_data_schema_are_changed_schema_version_is_bumped(init_data):
+    """
+    Test to validate that if the init data or run data schema are changed, we also bump the
+    integration_data_interface_version
+    """
+    # Define reference schemas that represent the "last known good" state
+    # When schemas change, these references AND the expected version should be updated together
+    REFERENCE_INIT_DATA_SCHEMA = {
```
MarinMaksutaj, line 26: Addressed feedback regarding test implementation:
- Removed the redundant schema loading that was comparing files against themselves
- Added reference schemas directly in the test file
- Test now properly compares current schemas against these references
- If schemas change from the reference, the test will fail unless the version is bumped

viknith (dismissed):

test/unit/deadline_adaptor_for_nuke/NukeAdaptor/test_schema_versioning.py
```diff
@@ -0,0 +1,83 @@
...
+    schemas_changed = (
+        current_init_data_schema != REFERENCE_INIT_DATA_SCHEMA
+        or current_run_data_schema != REFERENCE_RUN_DATA_SCHEMA
+    )
+
+    # If schemas have changed, version should be bumped
+    if schemas_changed:
+        assert semantic_version.major > EXPECTED_MAJOR or (
+            semantic_version.major == EXPECTED_MAJOR and semantic_version.minor > EXPECTED_MINOR
+        ), "If schemas change, integration_data_interface_version must be bumped"
```
joel-wong-aws: This test will only be useful at alerting us about schema version changes the first time the schema is modified (e.g. to 0.2). On subsequent schema modifications, this test will always pass, even if the schema version isn't increased. Would it simplify this test and increase test value by simply asserting that the reference data equals the expected data and giving a helpful error message if not?

MarinMaksutaj: Made some changes according to this comment.

joel-wong-aws (approved):

viknith (approved):

PR #200, opened 2025-03-13
fix: ensure continue_on_error is always passed to NukeHandler and set…

### LINKED ISSUE #198: Bug: Need the ability for the adaptor to not abort on errors (like Deadline 10)
### Describe the bug

Nuke graphs can contain 'errors' that are not detrimental to the output of a graph. 

The current adaptor will always exit on errors, but this requires all errors to be mitigated in the graph prior to submission - which is not how artist would like to work.

Deadline 10 has the ability to continue on errors.

Image

we would like an option ( like Deadline 10) to continue on errors( default is OFF)

### Expected Behaviour

add an option ( default = off) to continue on errors.

### Current Behaviour

adaptor will throw an error when Nuke encounters an error

### Reproduction Steps

I believe a file read node with a missing file will replicate the issue

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.)
2. Version of Nuke
3. Version of this package
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?

Please share other details about your environment that you think might be relevant to reproducing the bug.

MarinMaksutaj:

Artists need the ability to continue rendering even when Nuke encounters non-critical errors in the graph. Previously, any error in Nuke would cause the 
adaptor to abort the render, which is not always desirable. This behavior differs from Deadline 10, which has an option to continue on errors.

The requirement was to implement similar functionality in the Deadline Cloud for Nuke adaptor, allowing renders to continue despite encountering errors in 
the Nuke graph.

The solution implements a continue_on_error flag that is now always set to True by default. This is done by:

1. Hardcoding the continue_on_error property to return True (temporarily until we have a GUI option)
2. Adding the continue_on_error action at the beginning of the action queue
3. Enhancing the _handle_error method to always log errors but only raise exceptions if continue_on_error is False
4. Simplifying error handling in the on_run method

This change allows Nuke renders to continue even when encountering errors in the graph, which is a significant usability improvement for artists. They no 
longer need to fix all potential errors in their Nuke scripts before submission.

The impact is positive for users who want to continue rendering despite non-critical errors, which is a common workflow requirement.

The change was tested by:
1. Creating Nuke scripts with intentional errors (missing file read nodes)
2. Submitting these scripts to Deadline Cloud
3. Verifying that the render continues despite the errors
4. Checking that errors are properly logged but don't cause job failure

All tests passed successfully.

Yes, the change is documented in the code with comments explaining the behavior of continue_on_error. We will also update the user documentation to reflect
this new behavior.

No, this is not a breaking change. It enhances the functionality without breaking existing workflows. The default behavior is now more permissive (
continuing on errors), which aligns with user expectations from Deadline 10.

### DIFF
--- a/src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
+++ b/src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
@@ -9,7 +9,7 @@ import sys
 import threading
 import time
 import jsonschema  # type: ignore
-from typing import Callable, cast
+from typing import Callable
 
 from deadline.client.api import get_deadline_cloud_library_telemetry_client, TelemetryClient
 from openjd.adaptor_runtime._version import version as openjd_adaptor_version
@@ -111,7 +111,7 @@ class NukeAdaptor(Adaptor):
         Returns:
             bool: True if the behavior is to continue on errors. False otherwise.
         """
-        return cast(bool, self.init_data.get("continue_on_error", True))
+        return True  # Temporarily continue_on_error by default until we have an option for it on the GUI submitter
 
     @property
     def _nuke_is_running(self) -> bool:
@@ -219,6 +219,9 @@ class NukeAdaptor(Adaptor):
         Args:
             match (re.Match): The match object from the regex pattern that was matched the message
         """
+        # Always log the error
+        _logger.error(f"Nuke encountered an error: {match.group(0)}")
+
         if not self.continue_on_error:
             self._exc_info = RuntimeError(f"Nuke Encountered an Error: {match.group(0)}")
 
@@ -468,11 +471,14 @@ class NukeAdaptor(Adaptor):
         Client will request and perform. The action must be present in the _FIRST_NUKE_ACTIONS or
         _NUKE_INIT_KEYS set to be added to the action queue.
         """
+        # Always set continue_on_error to True first
+        self._action_queue.enqueue_action(Action("continue_on_error", {"continue_on_error": True}))
+
         for name in _FIRST_NUKE_ACTIONS:
             self._action_queue.enqueue_action(Action(name, {name: self.init_data[name]}))
 
         for name in _NUKE_INIT_KEYS:
-            if name in self.init_data:
+            if name in self.init_data and name != "continue_on_error":
                 self._action_queue.enqueue_action(Action(name, {name: self.init_data[name]}))
 
     def _get_deadline_telemetry_client(self):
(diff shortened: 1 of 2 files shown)


viknith (approved):

src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
```diff
@@ -111,7 +111,7 @@ def continue_on_error(self) -> bool:
         Returns:
             bool: True if the behavior is to continue on errors. False otherwise.
         """
-        return cast(bool, self.init_data.get("continue_on_error", True))
+        return True  # Temporarily continue_on_error by default until we have an option for it on the GUI submitter
```
epmog, line 114: Are we always sending `False` currently? This looks like a breaking change, and probably not something we want to do as a temporary measure. Have we looked at the amount of effort to add the toggle to the job settings? If it's already expose as a JobParameter in the template, it should be pretty simple

epmog (changes requested):

npmacl:
> The solution implements a continue_on_error flag that is now always set to True by default.

How did we decide which should be the default? I could see some people wanting this off, so that time isn't spent on a render that would come out incorrectly. How did we determine what the more common preference would be?

MarinMaksutaj:
> > The solution implements a continue_on_error flag that is now always set to True by default.
> 
> How did we decide which should be the default? I could see some people wanting this off, so that time isn't spent on a render that would come out incorrectly. How did we determine what the more common preference would be?

Good point, at the moment it's a bit ambiguous, this issue here https://github.com/aws-deadline/deadline-cloud-for-nuke/issues/198, makes it seem as if we might want to change the default too, but at the moment it is set as False in the template.

src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py
MarinMaksutaj: At this point I'm thinking it's better if I just rewrite this PR with the long term fix. As it is right now the customer has not contacted again about this issue, and I don't think they're aware of this branch. Making the GUI change should be quick. I'll add those changes in next revision.

MarinMaksutaj:
Update: As discussed above, the fix here would be to keep the default as False for `continue_on_error`, and allow for a toggle in the GUI submitter. Temporarily however, I'll leave this PR open, if the customer that is blocked wants to pull these changes to unblock themselves.

MarinMaksutaj:
Closing this as the long term fix was merged

PR #201, opened 2025-03-14
fix: Add timeout for Nuke jobs on EnvEnter/Exit

### LINKED ISSUE #194: Bug: missing adaptor environment enter and exit timeouts
### Describe the bug

The Nuke adaptor's environment enter and exits do not have a timeout specified in their [job template](https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/420638699a1ae5c7da30fbea2fc198194154c6e3/src/deadline/nuke_submitter/default_nuke_job_template.yaml#L98-L119).

### Expected Behaviour

The timeout for the environment enter should be a little more (ex. 1 minute) than the [server start timeout + Nuke start timeout](https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/420638699a1ae5c7da30fbea2fc198194154c6e3/src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py#L61-L64). The timeout for the environment exit should be a little more (ex. 1 minute) than the [server end timeout + Nuke end timeout](https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/420638699a1ae5c7da30fbea2fc198194154c6e3/src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py#L61-L64)

### Current Behaviour

If the environment enter or exit get stuck (and the first layer of timeouts doesn't work), then the worker agent does not enforce the timeout.

### Proposed Solution

Add a `timeout` to the environment enter and exit actions [here](https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/420638699a1ae5c7da30fbea2fc198194154c6e3/src/deadline/nuke_submitter/default_nuke_job_template.yaml#L98-L119), using the [OpenJD spec `timeout` field](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#5-action).

### Environment

1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.): affects all rendering OSes
2. Version of Nuke: affects all versions of Nuke
3. Version of this package: up to and including 0.18.8
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? N/A

Example PR: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/173

MarinMaksutaj:

The Nuke adaptor's environment enter and exit actions did not have timeout values specified in the job template. This could lead to situations where if the environment enter or exit gets stuck (and the first layer of timeouts doesn't work), the worker agent would not enforce any timeout.

Added timeout values of 600 seconds (10 minutes) to both the environment enter and exit actions in the Nuke job template. This provides enough time for the server and Nuke to start/stop with an additional buffer, similar to the approach used in other packages like Cinema 4D.

This change improves the reliability of Nuke jobs by ensuring that environment enter and exit actions have proper timeout handling. If these actions get stuck, the worker agent will now enforce the timeout and prevent jobs from hanging indefinitely.

The change was tested by running the formatting and linting checks, which passed successfully. The timeout values are consistent with the approach used in  other packages (e.g., Cinema 4D).

No additional documentation was needed as this is an internal implementation detail that doesn't affect the user-facing API or behavior. The timeout values are set to reasonable defaults that don't require user configuration.

No, this is not a breaking change. It adds timeout values where they were previously missing but doesn't change any existing functionality or interfaces.

### DIFF
--- a/src/deadline/nuke_submitter/default_nuke_job_template.yaml
+++ b/src/deadline/nuke_submitter/default_nuke_job_template.yaml
@@ -108,6 +108,7 @@ steps:
           - file://{{ Env.File.initData }}
           cancelation:
             mode: NOTIFY_THEN_TERMINATE
+          timeout: 600
         onExit:
           command: NukeAdaptor
           args:
@@ -117,6 +118,7 @@ steps:
           - '{{ Session.WorkingDirectory }}/connection.json'
           cancelation:
             mode: NOTIFY_THEN_TERMINATE
+          timeout: 600
   script:
     embeddedFiles:
     - name: runData


viknith (approved):

src/deadline/nuke_submitter/default_nuke_job_template.yaml
```diff
@@ -108,6 +108,7 @@ steps:
           - file://{{ Env.File.initData }}
           cancelation:
             mode: NOTIFY_THEN_TERMINATE
+          timeout: 600
```
epmog, line 111: How does this interact with the existing settings for Nuke timeouts? https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py#L59

If a user specifies no timeouts, do they now get a default of 10 minutes for the env enter? Related, but I believe if not specified then EnvExit will now have a default due to changes in OpenJobDescription.

npmacl, line 111: How long does this action take normally?

MarinMaksutaj: My thought was that by default now it would've been set to 10 minutes (time which I based it off from the other DCCs, for example Cinema4D). Anywhere else we are setting this value that I'm not aware?

MarinMaksutaj: 10 minutes seems like a generous number, not too sure how long it actually takes, but cinema4D has the same value specified for both actions.

MarinMaksutaj: I think I see what you mean now, there are default values on the GUI toggles themselves, and those would overwrite the default values on the template.. In that case we could always adjust those values to 10 minutes as well.

epmog: Definitely try out the existing timeout controls and see what behaviour we see when we change values, or choose to not use timeouts. What we end up doing should be clear and expected by the users.

eng-josh:
Closing this PR as stale.

PR #202, opened 2025-03-14
fix: Add default Linux host requirement for Nuke submitter

### LINKED ISSUE #190: Bug: Nuke submitter should have a default Linux host requirement
### Describe the bug

Currently, there are no default host requirements for the Nuke submitter. This means that when submitted to a queue connected to both Windows and Linux fleets, either could be assigned to the job. Generally, Windows workers are not used for Nuke renders and fail after picking up the job.

### Expected Behaviour

Nuke submissions should have a default host requirement of Linux only. Nuke is only supported on Linux on Deadline Cloud service managed fleets and is generally less expensive than on Windows for all fleet types.

### Current Behaviour

Windows workers may pick up Nuke jobs, causing those jobs to fail.

### Reproduction Steps

This affects any Nuke job submission.

As a workaround, customers can create separate a queue for Nuke jobs which is only connected to Linux fleets.

### Environment

1. Operating system: affects all submission operating systems
2. Version of Nuke: affects all versions of Nuke
3. Version of this package: affects all versions of this package, including 0.18.8
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? All versions of the submitter installer are affected, including 2025.01.16.0

MarinMaksutaj:
This change adds a default Linux host requirement for Nuke job submissions to prevent Windows workers from picking up Nuke jobs, which would fail. Nuke is only supported on Linux for Deadline Cloud service managed fleets and is generally less expensive than Windows for all fleet types.

Issue: #190

Currently, there are no default host requirements for the Nuke submitter. This means that when submitted to a queue connected to both Windows and Linux fleets, either could be assigned to the job. Windows workers are not used for Nuke renders and fail after picking up the job.

Added a default host requirement with Linux OS for the Nuke submitter by:
- Importing the necessary dataclasses (HostRequirements, OsRequirements)
- Creating a HostRequirements object with Linux OS requirement
- Passing this object to the SubmitJobToDeadlineDialog constructor

Nuke jobs will now default to requiring Linux hosts, preventing Windows workers from picking up these jobs and failing. This improves the user experience by reducing failed jobs and aligns with the fact that Nuke is only supported on Linux for Deadline Cloud service managed fleets.

Manually tested by submitting Nuke jobs and verifying that the Linux host requirement is correctly applied in the job submission dialog and in the resulting job bundle.

Not applicable for this change as it doesn't affect the job bundle structure, only adds a default value for an existing field.

The code includes comments explaining the rationale for the default Linux host requirement.

No, this is not a breaking change. It adds a sensible default that can still be overridden by users if needed.

Screenshot 2025-03-14 at 3 39 40 PM

### DIFF
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -12,6 +12,7 @@ import nuke
 from deadline.client.api import get_deadline_cloud_library_telemetry_client
 from deadline.client.job_bundle import deadline_yaml_dump
 from deadline.client.ui import gui_error_handler
+from deadline.client.ui.dataclasses import HostRequirements, OsRequirements
 from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import (  # type: ignore
     SubmitJobToDeadlineDialog,
     JobBundlePurpose,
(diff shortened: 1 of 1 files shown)


viknith (approved):

MarinMaksutaj:
Discussed offline, this would be a breaking change in certain scenarios where customers only have a queue with a windows fleet associated and we're setting the default to be Linux. For that reason I'll be closing this PR and the associated issue. If there's a need in the future we can always reopen. CC: @epmog

PR #203, opened 2025-03-14
fix: Add Windows support to GitHub Actions and fix path handling

### LINKED ISSUE #164: Maintenance: Need to add windows as supported github actions
### Description

Currently only linux and mac are part of supported github actions: https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/.github/workflows/code_quality.yml#L17

We would like to add windows as well. 

Currently, the following test case(`test_config_has_absolute_search_path`) does not work windows: https://github.com/diane-hj/deadline-cloud-for-nuke/blob/be4f3fc77bc0d0e8aa9c94e9239eaa29dbf17b95/test/unit/nuke_util/test_ocio.py#L112

The root cause could be because `PurePath` will evaluate to the current OS flavour, but these absolute posix paths don't look absolute on windows. 

```
>>> PurePath("/asdf/asdf")
PurePosixPath('/asdf/asdf')
>>> PurePath("/asdf/asdf").is_absolute()
True
>>> PureWindowsPath("/asdf/asdf").is_absolute()
False
```

### Solution

We need to fix the test case and enable windows into supported github actions.

MarinMaksutaj:
This change adds Windows as a supported platform in GitHub Actions workflows, and fixes the path handling in the OCIO module to correctly identify absolute paths across different operating systems.

Issue: #164

The GitHub Actions workflow was not running tests on Windows, and tests were failing on Windows because the OCIO module wasn't correctly identifying absolute paths on that platform. Specifically, POSIX-style absolute paths (like `/my/absolute/luts`) weren't being recognized as absolute on Windows.

1. Added Windows to the GitHub Actions workflow matrix by adding `windows-latest` to the list of supported operating systems.
2. Modified the `config_has_absolute_search_paths` function to check paths using both the native path handling (via `PurePath`) and POSIX path handling (via `PurePosixPath`) to ensure cross-platform compatibility.

- The GitHub Actions workflow will now run tests on Windows, providing better test coverage across platforms.
- The OCIO module will correctly identify absolute paths on all platforms, including Windows.
- This fixes the test failures on Windows, allowing for better cross-platform support.

The changes were tested by running the specific test that was failing (`test_config_has_absolute_search_paths`) locally, which now passes. The full test suite will be run by the GitHub Actions workflow on all platforms, including Windows.

Not applicable for this change as it only affects the test infrastructure and path handling logic, not the job bundle output.

The code changes include clear comments explaining the cross-platform path handling approach.

No, this is not a breaking change. It enhances the existing functionality to work correctly across all platforms.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,19 +116,19 @@ line-length = 100
 xfail_strict = true
 addopts = [
     "--durations=5",
+    "--color=yes",
     "--cov=src/deadline/nuke_adaptor",
     "--cov=src/deadline/nuke_util",
     "--cov=src/deadline/nuke_submitter",
-    "--color=yes",
     "--cov-report=html:build/coverage",
     "--cov-report=xml:build/coverage/coverage.xml",
     "--cov-report=term-missing",
-    "--numprocesses=auto",
+    "--numprocesses=auto"
 ]
-testpaths = [ "test/unit" ]
+testpaths = [ "test" ]
 looponfailroots = [
     "src",
-    "test/unit",
+    "test",
 ]
 # looponfailroots is deprecated, this removes the deprecation from the test output
 filterwarnings = [
--- a/src/deadline/nuke_util/ocio.py
+++ b/src/deadline/nuke_util/ocio.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import os
-from pathlib import PurePath
+from pathlib import PurePath, PurePosixPath
 from typing import Optional
 
 import nuke
@@ -58,7 +58,17 @@ def create_config_from_file(ocio_config_path: str) -> OCIO.Config:
 
 def config_has_absolute_search_paths(ocio_config: OCIO.Config) -> bool:
     """True if any paths in the OCIO config's search path are absolute"""
-    return any(PurePath(path).is_absolute() for path in ocio_config.getSearchPaths())
+    # Check each path in the search paths
+    for path in ocio_config.getSearchPaths():
+        # Check if it's absolute according to the current OS
+        if PurePath(path).is_absolute():
+            return True
+
+        # Also check if it's a POSIX absolute path (for cross-platform compatibility)
+        if PurePosixPath(path).is_absolute():
+            return True
+
+    return False
 
 
 def get_config_absolute_search_paths(ocio_config: str | OCIO.Config) -> list[str]:
--- a/src/deadline/nuke_util/ocio.py
+++ b/src/deadline/nuke_util/ocio.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import os
-from pathlib import PurePath, PurePosixPath
+from pathlib import PurePath
 from typing import Optional
 
 import nuke
@@ -58,17 +58,7 @@ def create_config_from_file(ocio_config_path: str) -> OCIO.Config:
 
 def config_has_absolute_search_paths(ocio_config: OCIO.Config) -> bool:
     """True if any paths in the OCIO config's search path are absolute"""
-    # Check each path in the search paths
-    for path in ocio_config.getSearchPaths():
-        # Check if it's absolute according to the current OS
-        if PurePath(path).is_absolute():
-            return True
-
-        # Also check if it's a POSIX absolute path (for cross-platform compatibility)
-        if PurePosixPath(path).is_absolute():
-            return True
-
-    return False
+    return any(PurePath(path).is_absolute() for path in ocio_config.getSearchPaths())
 
 
 def get_config_absolute_search_paths(ocio_config: str | OCIO.Config) -> list[str]:
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -14,7 +14,7 @@ jobs:
     name: Python
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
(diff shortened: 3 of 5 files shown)


viknith:
Can you trigger Github Actions to verify the test will pass there? Or was it always faling locally as well?

pyproject.toml
```diff
@@ -116,19 +116,19 @@ line-length = 100
...
     "--cov=src/deadline/nuke_submitter",
-    "--color=yes",
     "--cov-report=html:build/coverage",
     "--cov-report=xml:build/coverage/coverage.xml",
     "--cov-report=term-missing",
-    "--numprocesses=auto",
+    "--numprocesses=auto"
 ]
-testpaths = [ "test/unit" ]
+testpaths = [ "test" ]
```
viknith, line 128: Any reason not to specify unit?

viknith:
What was the user experience before the paths issue was fixed, any different from how it looks now? Is the change just on the SMF rendering side or does it impact anything on the submitter side?

pyproject.toml
MarinMaksutaj: This was another problem with windows tests. Apparently windows detects paths slightly differently, which cause no tests to run when we had tests/unit.. going up one level fixed the issue.

MarinMaksutaj:
> Can you trigger Github Actions to verify the test will pass there? Or was it always faling locally as well?

Yes, tests have already been triggered, look at the `Checks` tab for the new windows Github action run.

MarinMaksutaj:
> What was the user experience before the paths issue was not fixed, any different from how it looks now? Is the change just on the SMF rendering side or does it impact anything on the submitter side?

Total failure of windows tests for all 3 python versions.

src/deadline/nuke_util/ocio.py
```diff
@@ -58,7 +58,17 @@ def create_config_from_file(ocio_config_path: str) -> OCIO.Config:
 
 def config_has_absolute_search_paths(ocio_config: OCIO.Config) -> bool:
     """True if any paths in the OCIO config's search path are absolute"""
-    return any(PurePath(path).is_absolute() for path in ocio_config.getSearchPaths())
```
viknith: Nit: could still be a one-liner and do `and PurePosixPath(path).is_absolute()` here, yes?

viknith (dismissed):

src/deadline/nuke_util/ocio.py
```diff
@@ -58,7 +58,17 @@ def create_config_from_file(ocio_config_path: str) -> OCIO.Config:
...
+    # Check each path in the search paths
+    for path in ocio_config.getSearchPaths():
+        # Check if it's absolute according to the current OS
+        if PurePath(path).is_absolute():
+            return True
+
+        # Also check if it's a POSIX absolute path (for cross-platform compatibility)
+        if PurePosixPath(path).is_absolute():
+            return True
```
epmog: I'm confused by _why_ this was added. Windows paths are not PosixPaths, so I don't think this is doing what you want it to.

PurePath will choose PurePosixPath vs. PureWindowsPath based on your OS: https://docs.python.org/3/library/pathlib.html#pure-paths. So PurePath will work on Windows and Linux (+macOS) when ran on those operating systems respectively: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_absolute. 

Have you checked the results you're getting when Windows paths are being fed in here when ran on Windows? If you're running this on a posix machine, then yes the Windows paths will appear to not be absolute when using PurePath.

epmog (changes requested):

src/deadline/nuke_util/ocio.py
MarinMaksutaj: The issue we're addressing (#164) involves handling absolute paths in OCIO configs across different platforms. The challenge is that OCIO configs often contain POSIX-style paths (starting with '/') regardless of the platform they're used on.

From the Issue:

```
>>> PurePath("/asdf/asdf")
PurePosixPath('/asdf/asdf')
>>> PurePath("/asdf/asdf").is_absolute()
True
>>> PureWindowsPath("/asdf/asdf").is_absolute()
False
```

Consider this scenario:
1. A user creates an OCIO config on Linux with paths like `/test/my_test.cube`
2. When this config is used on Windows, these paths are still POSIX-style
3. Using only `PurePath().is_absolute()` on Windows would incorrectly identify these as non-absolute paths, since Windows expects paths like `C:\test\my_test.cube`

The solution needs to handle both:
- Native Windows paths (using `PurePath` which resolves to `PureWindowsPath` on Windows)
- POSIX-style paths that might exist in OCIO configs created on Linux/Mac

epmog: The linked issue isn't clear to me that Nuke is doing `/posix/rooted` paths on Windows. It looks like we have tests that mock out Nuke that had previously defined posix paths which didn't work when adding windows test runners because those are correctly identified as not absolute.

For the example scenario, if we're grabbing OCIO configs on one platform and trying to use them on another, then we should be adding the correct platform path.

> /test/my_test.cube
> ...
> Using only PurePath().is_absolute() on Windows would incorrectly identify these as non-absolute paths

That is Windows behaving correctly. That is a relative path on Windows, but also interacts with the current drive selected.

If we're grabbing Linux paths and trying to use them on Windows, then we should be applying path mapping or setting the config so that we're referencing the correct values.

MarinMaksutaj:
Created another PR to only enable windows github action workflow without fixing ocio tests.

PR #206, opened 2025-03-21
Nuke individual installer setup

MarinMaksutaj:

We needed to create an individual installer specifically for Nuke integration with Deadline Cloud. This allows users to install the Deadline Cloud for Nuke plugin separately from other integrations, providing a more streamlined and focused installation experience for Nuke users.

Created a dedicated installer setup for the Nuke integration by:
• Adding installer configuration files specific to Nuke
• Creating build scripts to generate the installer
• Adding necessary assets (icons, splash screens, etc.)
• Implementing installer-specific tests to ensure functionality
• Setting up proper naming conventions for the Nuke-specific installer

Users can now install the Deadline Cloud for Nuke integration separately, without needing to install other integrations. This improves the user experience for Nuke-specific workflows and reduces installation footprint for users who only need Nuke integration.

The changes were tested by:
• Running unit tests for the installer scripts
• Building and verifying the installer package
• Testing the installation process on supported platforms
• Verifying the Nuke integration works correctly after installation

Yes, the installer includes appropriate documentation and the build scripts contain comments explaining the process. The installer itself provides user guidance during the installation process.

No, this is an additive change that provides a new installation option without affecting existing functionality.


installer/DeadlineCloudForBlenderSubmitter.xml
```diff
@@ -0,0 +1,243 @@
+<project>
```
viknith: Probably should remove this Blender one

MarinMaksutaj: Removed. Thanks for the catch!

.github/workflows/build_installers.yml
```diff
@@ -0,0 +1,34 @@
+name: Build Installers
+
```
viknith, line 2: I see this template came from [here](https://github.com/aws-deadline/deadline-cloud-for-blender/pull/208/files#diff-acf7eb350af5eb781c183052ab7307bb58fa6221fb0a36e6784476941f1cb16b) and the template I referenced came from [here](https://github.com/aws-deadline/.github/pull/32/files). I wonder if there's much difference

hatch.toml
```diff
@@ -6,6 +6,7 @@ pre-install-commands = [
 [envs.default.scripts]
 sync = "pip install -r requirements-testing.txt"
 test = "pytest {args:test/unit/*}"
+test-installer = "pytest --no-cov {args:test/installer} -vvv --numprocesses=1"
```
viknith: Any reason for numprocesses to be 1?

MarinMaksutaj: Most likely left out from testing locally, will make sure to remove it with another commit once we also get more information on the Github workflow issue.

.github/workflows/build_installers.yml
viknith: No issue, we're matching on this now.

installer/DeadlineCloudForNuke.xml
```diff
@@ -0,0 +1,243 @@
...
+            <actionList>
+                <showWarning text="Cannot uninstall a system-wide installation as a non-administrator user. Please run the uninstaller as an administrator."/>
+                <exit exitCode="1"/>
+            </actionList>
+            <ruleList>
+                <compareText text="system" logic="equals" value="${installscope}"/>
+                <isFalse value="${installer_is_root_install}"/>
+            </ruleList>
+        </actionGroup>
+        <deleteFile path="${installdir}/installer_version.txt"/>
```
viknith, line 240: Nit: probably don't need this, but not blocking (same issue in AE too)

viknith (dismissed):

viknith (approved):

karthikbekalp (approved):

PR #208, opened 2025-04-07
fix: add continue on error option to submitter UI

MarinMaksutaj:

Nuke graphs can contain 'errors' that are not detrimental to the output of a graph. The current adaptor was always exiting on errors, requiring all errors  to be mitigated in the graph prior to submission - which is not how artists prefer to work. Deadline 10 has the ability to continue on errors, and we  needed to provide similar functionality in Deadline Cloud for Nuke.

Added a "Continue on error" checkbox to the submitter UI that allows users to control whether Nuke should continue rendering when it encounters non-fatal errors in the graph. This implementation:
1. Added a continue_on_error field to the RenderSubmitterUISettings class
2. Added a checkbox to the SceneSettingsWidget UI with a helpful tooltip
3. Added code to pass the setting to the job template parameters

Artists can now choose whether they want Nuke to continue rendering when it encounters non-fatal errors in the graph, providing a more flexible workflow similar to what was available in Deadline 10.

The changes were tested by manually verifying that:
1. The checkbox appears in the UI
2. The setting is saved as part of the sticky settings
3. The value is correctly passed to the job template parameters

Yes, the checkbox has a tooltip explaining its purpose: "Allow Nuke to continue rendering when it encounters non-fatal errors in the graph"

No, this is not a breaking change. It adds a new option that defaults to false, maintaining the current behavior while providing additional functionality.

image

### DIFF
--- a/src/deadline/nuke_submitter/data_classes.py
+++ b/src/deadline/nuke_submitter/data_classes.py
@@ -26,6 +26,7 @@ class RenderSubmitterUISettings:  # pylint: disable=too-many-instance-attributes
     write_node_selection: str = field(default="", metadata={"sticky": True})
     view_selection: str = field(default="", metadata={"sticky": True})
     is_proxy_mode: bool = field(default=False, metadata={"sticky": True})
+    continue_on_error: bool = field(default=False, metadata={"sticky": True})
 
     input_filenames: list[str] = field(default_factory=list, metadata={"sticky": True})
     input_directories: list[str] = field(default_factory=list, metadata={"sticky": True})
--- a/src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
+++ b/src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
@@ -65,7 +65,7 @@ class SceneSettingsWidget(QWidget):
         self.continue_on_error_check.setToolTip(
             "Allow Nuke to continue rendering when it encounters non-fatal errors in the graph"
         )
-        lyt.addWidget(self.continue_on_error_check, 3, 1)
+        lyt.addWidget(self.continue_on_error_check, 4, 0)
 
         self.timeout_checkbox = QCheckBox("Use timeouts", self)
         self.timeout_checkbox.setChecked(True)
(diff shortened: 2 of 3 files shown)


src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
```diff
@@ -61,23 +61,29 @@ def _build_ui(self):
...
+        )
+        lyt.addWidget(self.continue_on_error_check, 4, 0)
+
         self.timeout_checkbox = QCheckBox("Use timeouts", self)
         self.timeout_checkbox.setChecked(True)
         self.timeout_checkbox.clicked.connect(self.activate_timeout_changed)
         self.timeout_checkbox.setToolTip(
             "Set a maximum duration for actions from this job. See AWS Deadline Cloud documentation to learn more"
         )
-        lyt.addWidget(self.timeout_checkbox, 4, 0)
```
viknith, line 70: Can you elaborate on what these number changes are doing?

MarinMaksutaj: They represent the row position in the GUI window, since we're adding a new checkbox, shifting everything down by 1. Posted a screenshot to help visualize the changes.

viknith (approved):

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
npmacl: If Qt doesn't have it already (I assume this is qt) I think it would be good in the future to refactor and abstract this so that we don't need to manage this manually.

src/deadline/nuke_submitter/data_classes.py
```diff
@@ -26,6 +26,7 @@ class RenderSubmitterUISettings:  # pylint: disable=too-many-instance-attributes
     write_node_selection: str = field(default="", metadata={"sticky": True})
     view_selection: str = field(default="", metadata={"sticky": True})
     is_proxy_mode: bool = field(default=False, metadata={"sticky": True})
+    continue_on_error: bool = field(default=False, metadata={"sticky": True})
```
npmacl, line 29: I assume what sticky here means is that this will be False by default, but if we change it to true it will remember that for subsequent submissions? 

If I close and reopen nuke, is it going to remember what I had selected last time? If not I feel like there would be some inconvenience where you get into the muscle memory of assuming this option is what you wanted, and then when you open it the next day you forget to set it and submit a job that fails. I still think this is better than having to click it every time, but If it's possible to have this be remembered I think that would be ideal.

MarinMaksutaj: That's actually correct, sticky meaning that if we check the box, we can close and open the Nuke application, and that option will be remembered on the subsequent submission.

npmacl: awesome

npmacl (approved):

PR #219, opened 2025-05-12
Add support for Nuke 16

MarinMaksutaj:

The only change we need to perform to be compatible with Nuke 16 would be using PySide 6 instead of version 2 that we use for Nuke 14 and Nuke 15.

Added logic to default to PySide6 library only when Nuke version is 16, and falls back to our default versions for 14-15 versions.

Add support for Nuke 16

Tested it by running the submitter for Nuke 16.

Yes, changed the corresponding values in XML file to display Nuke 16 support

Should not be since we fall back to default versions for every other non Nuke 16 version.

### DIFF
--- a/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
+++ b/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
@@ -18,12 +18,29 @@ from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import (  # type: 
 )
 from deadline.nuke_util import ocio as nuke_ocio
 from nuke import Node
-from PySide2.QtCore import Qt  # pylint: disable=import-error
-from PySide2.QtWidgets import (  # pylint: disable=import-error; type: ignore
-    QApplication,
-    QMainWindow,
-    QMessageBox,
-)
+
+# Handle different Qt imports for different Nuke versions
+try:
+    # For Nuke 16+
+    from PySide6.QtCore import Qt
+    from PySide6.QtWidgets import QApplication, QMainWindow, QMessageBox
+except ImportError:
+    try:
+        # For Nuke 13-15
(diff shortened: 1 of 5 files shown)


viknith (approved):

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -18,12 +18,20 @@
...
-    QApplication,
-    QMainWindow,
-    QMessageBox,
-)
+
+# Handle different Qt imports for different Nuke versions
+try:
+    # For Nuke 16+
+    from PySide6.QtCore import Qt
+    from PySide6.QtWidgets import QApplication, QMainWindow, QMessageBox
```
eherozhao, line 26: nit: can we add a log to trace this? also can our telemetry identify if customers is using 16 or other version?

MarinMaksutaj: We do through `get_deadline_cloud_library_telemetry_client`:

```
# In show_nuke_render_submitter function
get_deadline_cloud_library_telemetry_client().update_common_details({
    "deadline-cloud-for-nuke-submitter-version": version,
    "nuke-version": nuke.env["NukeVersionString"],
})
```

The available Nuke version variables are:
• `nuke.env["NukeVersionString"]` - Full version string (e.g., "16.0v1")
• `nuke.env["NukeVersionMajor"]` - Major version number (e.g., 16)
• `nuke.env["NukeVersionMinor"]` - Minor version number
• `nuke.env["NukeVersionRelease"]` - Release version number


As for the log suggestion, I'd say not urgently needed here, we will eventually crash if there's an import issue and the error trace will show in the terminal that opens Nuke. We can add it later if needed but I don't see it as blocking.

eherozhao (approved):

PR #226, opened 2025-06-20
feat: Nuke Submitter E2E Testing: Implemented the basic workflow validation test with Squish GUI tests, API verification, and output validation

RenaL27:
Enhanced the test automation framework to replicate the basic Nuke job submission workflow:
- Uses Squish to automate UI interactions for job submission, replicating customer workflow
- Leverages Deadline API to verify job status, farm/queue/storage settings
- Implements statistical RGB comparison to validate rendered image sequences
- Adds systematic cleanup of test outputs

The framework currently implements the Basic Job Submission test case, ensuring reliable end-to-end testing of the core submission workflow. **The Basic Submission Workflow currently only supports Mac. Support for other operating systems will come soon.**

More reliable test automation for Nuke Submitter through E2E testing. This end to end testing was added to the Squish folder within the test directory. 

The basic submission workflow was tested through running the hatch command `hatch run squish:test`, which triggers the `test_basic_workflow` pytest. This test will launch Squish and Nuke, submits a job, verifies it's success and job parameters using the Deadline Python library. Then, it will compare image outputs between the expected and the output. Lastly it will cleanup the resources by deleting the image outputs that were generating.

Since these changes are focused on the test automation framework and don't modify the core job bundle functionality, running the Job Bundle Output Tests is not required. The changes are isolated to the test framework implementation.
```
```

Yes, the changes are through clear function and parameter documentation, logging, error messages, and more.

No, this is not a breaking change. The modifications are isolated to the test automation framework and don't affect the core functionality of the Nuke Submitter or its interfaces.
----

### DIFF
--- a/test/squish/SQUISH_README.md
+++ b/test/squish/SQUISH_README.md
@@ -30,7 +30,7 @@ NUKE_PATH=/Users/user/DeadlineCloudForNukeSubmitter
 ### Set Required Environment Variables
 The following environment variables are required for running the tests:
 ```sh
-export AWS_PROFILE=<ada profile name>
+export AWS_PROFILE=<profile name>
 export DEADLINE_NUKE_PATH=/path/to/deadline-cloud-for-nuke
 export NUKE_ASSET_ROOT=/path/to/nuke/assets
 ```
@@ -53,9 +53,29 @@ The following Deadline Cloud resources are needed in order to run `tst_verify_se
 
 ## Available Tests
 ### Basic Workflow Test
-The `basic_workflow` test provides comprehensive end-to-end validation of the basic Nuke submitter workflow. This test consists of two main parts: First, the basic_workflow_gui Squish test automates the UI interaction by launching Nuke, opening a pre-configured script file with a write node, configuring AWS profile and resources, submitting a job to Deadline Cloud, and closing Nuke. Second, the test performs backend validation by verifying the job was successfully submitted to the correct queue, monitoring job completion, downloading the rendered output files, comparing them against reference images to ensure visual accuracy, and handling the cleanup of resourcse.
+The `basic_workflow` test provides comprehensive end-to-end validation of the basic Nuke submitter workflow. This test consists of two main parts: First, the basic_workflow_gui Squish test automates the UI interaction by launching Nuke, opening a pre-configured script file with a write node, configuring AWS profile and resources, submitting a job to Deadline Cloud, and closing Nuke. Second, the test performs backend validation by verifying the job was successfully submitted to the correct queue, monitoring job completion, downloading the rendered output files, comparing them against reference images to ensure visual accuracy, and handling the cleanup of resource.
+
 
 ## Running Tests
 ```sh
 hatch run squish:test
 ```
+
+## Test Results Interpretation
+### Successful vs. Unsuccessful Tests
+A successful test will be indicated by pytest's green message saying that the test has passed.
+
+An unsuccessful test will show:
+- FAIL or FATAL messages in the Squish output
+- Error messages like "Job ID from Squish execution does not match the latest job ID"
+- Assertion failures for squish commands, job submission, download, or job verification steps
+
+### Image Verification Failures
+When image verification fails, you can check the output images located at `$NUKE_ASSET_ROOT/nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/images/output` 
+
+These output images can be compared with the expected reference images at:
+`$NUKE_ASSET_ROOT/nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/images/expected`.
+
+The test compares RGB values across the frames and calculates an average difference. A difference exceeding the tolerance (default: 0.1) will cause the test to fail.
+
+In the case changes are made to the Nuke script, such as adding color correction nodes, applying visual effects, or modifying render settings, the output images will likely differ from the reference images. When these changes are intentional, you should update the reference images to reflect the new expected output.
--- a/test/squish/SQUISH_README.md
+++ b/test/squish/SQUISH_README.md
@@ -1,0 +1,61 @@
+# Nuke Submitter E2E Testing with Squish
+
+Nuke and Squish require a license. If you have a Squish and Nuke license, please follow the guide below to run the tests.
+
+## Prerequisites
+### Install Nuke
+
+Download and install Nuke 16.0v1. The current tests have been validated on macOS 15.5. Support for Windows and Linux will be added after macOS support is established. 
+
+### Set up the Deadline Cloud Nuke Submitter
+
+Read the DEVELOPMENT.md for instructions on setting up the Nuke submitter.
+
+### Install Squish Framework
+
+Install Squish 8.1.0 for Qt 6.5. If you are using any other version, be sure to select the correct version of Qt that is being used with Nuke on your machine.
+
+## Configure Squish Environment
+
+Register Nuke as an AUT (Application Under Test) by going to 'Edit' -> 'Server Settings' and registering under 'Mapped AUTs' (in Squish IDE). 
+
+Then, configure the Nuke Submitter path by going to the test suite settings in the Squish IDE and adding an AUT environment variable called NUKE_PATH. Set it to the path where your Nuke submitter is installed: 
+```sh
+/path/to/DeadlineCloudForNukeSubmitter
+```
+You can also modify the envvars file:
+```sh
+NUKE_PATH=/Users/user/DeadlineCloudForNukeSubmitter
+```
+### Set Required Environment Variables
+The following environment variables are required for running the tests:
+```sh
+export AWS_PROFILE=<ada profile name>
+export DEADLINE_NUKE_PATH=/path/to/deadline-cloud-for-nuke
+export NUKE_ASSET_ROOT=/path/to/nuke/assets
+```
+
+### AWS Authentication
+Paste your temporary AWS credentials in the terminal session where you'll run the tests. These credentials should have permissions to access the Deadline Cloud resources.
+
+### Deadline Cloud Monitor
+Log in to the Deadline Cloud Monitor to verify your credentials are working and that you can access the required resources.
+
+## Deadline Cloud Resources Needed for Running Tests
+
+The following Deadline Cloud resources are needed in order to run `tst_verify_settings_dialogue` test suite:
+
+- An AWS default profile (used for authentication)
+- A farm named "Nuke Submitter Squish Farm"
+- A queue named "Nuke Submitter Squish Automation Queue"
+- Three storage profiles named "Linux Storage Profile", "Windows Storage Profile", and "macOS Storage Profile"
+- Fleet instance market type of On-Demand instance
+
+## Available Tests
+### Basic Workflow Test
+The `basic_workflow` test provides comprehensive end-to-end validation of the basic Nuke submitter workflow. This test consists of two main parts: First, the basic_workflow_gui Squish test automates the UI interaction by launching Nuke, opening a pre-configured script file with a write node, configuring AWS profile and resources, submitting a job to Deadline Cloud, and closing Nuke. Second, the test performs backend validation by verifying the job was successfully submitted to the correct queue, monitoring job completion, downloading the rendered output files, comparing them against reference images to ensure visual accuracy, and handling the cleanup of resourcse.
+
+## Running Tests
+```sh
+hatch run squish:test
+```
--- a/test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
+++ b/test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
@@ -1,35 +1,42 @@
 # -*- coding: utf-8 -*-
-
+# mypy: disable-error-code="attr-defined"
 import names
 import squish_helpers
 import os
 import test
 import squish
 
+
 def main():
     squish_helpers.launch_nuke()
-    squish.setWindowState(squish.waitForObject(names.nukeMainWindow_Foundry_UI_DockMainWindow), squish.WindowState.Maximize)
-    
+    squish.setWindowState(
+        squish.waitForObject(names.nukeMainWindow_Foundry_UI_DockMainWindow),
+        squish.WindowState.Maximize,
+    )
+
     # Open the nuke script file that is being tested
     squish.activateItem(squish.waitForObjectItem(names.o_QMenuBar, "File"))
     squish.activateItem(squish.waitForObjectItem(names.file_Foundry_UI_Menu, "Open Comp..."))
-    
+
     file_path_edit = squish.waitForObject(names.script_to_open_FilePathEdit)
 
-    nuke_asset_path = os.environ.get('NUKE_ASSET_ROOT')
+    nuke_asset_path = os.environ.get("NUKE_ASSET_ROOT")
     if not nuke_asset_path:
         test.fatal("NUKE_ASSET_ROOT environment variable not set")
         return
-    
-    script_path = os.path.join(nuke_asset_path, "nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/scripts/nukeSubmitter_v02_nuke_default_one_write_node.nk")
+
+    script_path = os.path.join(
+        nuke_asset_path,
+        "nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/scripts/nukeSubmitter_v02_nuke_default_one_write_node.nk",
+    )
     test.log(f"Using script path: {script_path}")
-    
+
     squish.setFocus(file_path_edit)
     squish.type(file_path_edit, script_path)
-    
+
     # Opens the nuke script file
     squish.clickButton(squish.waitForObject(names.script_to_open_Open_QPushButton))
-    
+
     squish.type(squish.waitForObject(names.nukeMainWindow_DAG_DAG_Window), "<Ctrl+S>")
     squish_helpers.open_nuke_submitter_gui()
     squish_helpers.configure_aws_profile()
--- a/test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
+++ b/test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
@@ -34,11 +34,8 @@ def main():
         test.fatal("NUKE_ASSET_ROOT environment variable not set")
         return
 
-    script_path = os.path.join(
-        nuke_asset_path,
-        "nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/scripts/nukeSubmitter_v02_nuke_default_one_write_node.nk",
-    )
-    test.log(f"Using script path: {script_path}")
+    NUKE_TEST_SCRIPT_PATH = "nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/scripts/nukeSubmitter_v02_nuke_default_one_write_node.nk"
+    script_path = os.path.join(nuke_asset_path, NUKE_TEST_SCRIPT_PATH)
 
     squish.setFocus(file_path_edit)
     squish.type(file_path_edit, script_path)
--- a/test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
+++ b/test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+
+import names
+import squish_helpers
+import os
+import test
+import squish
+
+def main():
+    squish_helpers.launch_nuke()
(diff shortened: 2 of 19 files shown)


test/squish/suite_nuke_submitter/envvars
```diff
@@ -0,0 +1 @@
+NUKE_PATH=/Users/renali/DeadlineCloudForNukeSubmitter
```
viknith: This should be updated to something more generic, or add docs on what a user should do to set this and gitignore it to prevent pushing this up to Github perhaps.

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
```diff
@@ -0,0 +1,44 @@
...
+    nuke_asset_path = os.environ.get("NUKE_ASSET_ROOT")
+    if not nuke_asset_path:
+        test.fatal("NUKE_ASSET_ROOT environment variable not set")
+        return
+
+    script_path = os.path.join(
+        nuke_asset_path,
+        "nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/scripts/nukeSubmitter_v02_nuke_default_one_write_node.nk",
+    )
+    test.log(f"Using script path: {script_path}")
```
viknith: I saw print in other places and test.log here, what's the difference?

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+        return result.returncode == 0, job_id
+    except Exception as e:
+        print(f"Error running squish command: {e}")
+        return False
+
+
+def test_basic_workflow():
+    deadline_nuke_path = os.environ.get("DEADLINE_NUKE_PATH")
+    squish_command = f"/Applications/Squish\\ for\\ Qt\\ 8.1.0/bin/squishrunner --testsuite {deadline_nuke_path}/test/squish/suite_nuke_submitter --testcase basic_workflow_gui --local"
```
viknith: I'm guessing `/Applications/Squish\\ for\\ Qt\\ 8.1.0/bin/squishrunner --testsuite {deadline_nuke_path}/test/squish/suite_nuke_submitter` will always be a part of the command being run. Should we move this into the run_squish_command function (maybe rename to run_squish_testcase) and here we would just pass in "basic_workflow_gui" as the input parameter? Or does the prefix change a lot depending on the test case?

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+# mypy: disable-error-code="attr-defined"
+import names
+import squish_helpers
+import os
+import test
+import squish
+
+
+def main():
```
viknith, line 20: Can we get some more comments here to understand what this file does and what triggers it?

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+
+from shared.scripts import config
+from deadline.client.api import get_boto3_client, list_farms, list_queues, list_jobs
+from botocore.exceptions import ClientError
+from deadline.job_attachments.download import OutputDownloader
+from deadline.job_attachments.models import JobAttachmentS3Settings
+import time
+
+
+def get_farm_id_by_name():
```
viknith, line 11: Can't you just get the farm ID from the deadline CLI default farm_id that's set? Same question for queues. I'm not sure how farm_name comes into play.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+        storage_profile = deadline.get_storage_profile(
+            farmId=farm_id, storageProfileId=storage_profile_id
+        )
+
+        # Assert it's the macOS profile
+        print(storage_profile["osFamily"])
+        assert storage_profile["osFamily"] == "MACOS", "Storage profile is not macOS"
+        print(f"Verified storage profile is macOS: {storage_profile['displayName']}")
+
+        print("Verified that default farm/queue/storage is selected")
```
viknith, line 72: Why does it need to have a storage profile that's a mac OS storage profile? IIRC that shouldn't matter, unless you're doing testing that involves storage profiles?

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+        time.sleep(poll_interval)
+
+
+def download_output(farm_id=None, queue_id=None, job_id=None):
+    """Download job outputs after verifying job completion."""
+    try:
+        # Wait for job completion
+        job_complete, job_status = wait_for_job_completion(farm_id, queue_id, job_id)
+        if not job_complete or job_status != "SUCCEEDED":
+            return False, f"Job did not complete successfully: {job_status}"
```
viknith: The second return value here seems to be more appropriate as a log statement rather than being returned, unless the message has a use case. Same thing for all other instances of this.

test/squish/suite_nuke_submitter/shared/scripts/cleanup_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+    try:
+        deleted_count = 0
+        output_path = Path(output_dir)
+
+        # Verify output directory exists
+        if not output_path.exists():
+            return True, f"Output directory {output_dir} does not exist, nothing to clean"
+
+        # Delete each frame
+        for frame in range(start_frame, end_frame + 1):
```
viknith: Nit: wouldn't it be simpler to just iterate through all the files and if the file has prefix `base_name`, then delete it? Or are there any risks associated with that?

test/squish/suite_nuke_submitter/shared/scripts/names.py
```diff
@@ -0,0 +1,314 @@
+# encoding: UTF-8
+
+# from objectmaphelper import *
+
+nukeMainWindow_Foundry_UI_DockMainWindow = {
```
viknith, line 57: What is this file for?

test/squish/suite_nuke_submitter/shared/scripts/squish_helpers.py
```diff
@@ -0,0 +1,155 @@
+# -*- coding: utf-8 -*-
+# mypy: disable-error-code="attr-defined"
+import squish
+import test
+import names
+import config
+
```
viknith, line 7: What happens if any of these helper function fail because the user didn't install Nuke or Squish correctly? Do we need to add any exception handling to notify the user in logs about that or is it already visible?

test/squish/suite_nuke_submitter/shared/scripts/squish_helpers.py
```diff
@@ -0,0 +1,155 @@
...
+def configure_aws_profile():
+    squish.clickButton(
+        squish.waitForObject(names.submit_to_AWS_Deadline_Cloud_Settings_QPushButton)
+    )
+    squish.sendEvent(
+        "QMoveEvent",
+        squish.waitForObject(
+            names.aWS_Deadline_Cloud_workstation_configuration_DeadlineConfigDialog
+        ),
+        541,
```
viknith, line 69: Are these numbers coordinates on the screen based on pixel count? If so, how does this differ on larger or smaller screens? Do we need some custom sizing logic to make sure Squish clicks the right location on screen?

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+    except ClientError as e:
+        error_code = e.response["Error"]["Code"]
+        if error_code == "ResourceNotFoundException":
+            print(f"Job not found with ID: {job_id}")
+        else:
+            print(f"Error getting job: {e}")
+        return None
+
+
+def verify_job_in_queue(farm_id=None, queue_id=None, job_id=None):
```
viknith: Do all of these parameters need these None default values? Same question for all instances of this.

viknith (commented):
Can you add/update the DEVELOPMENT.md or add a new README.md file in here to explain the anatomy of these files and how everything interacts at a high-level?

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
RenaL27: Yes, will do!

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
RenaL27: test.log is specifically used for Squish test script reporting since it appears in Squish test results/reports while print is used in the Python scripts that run those Squish tests.

test/squish/suite_nuke_submitter/run_squish_test.py
RenaL27: Yes, moving this command into the run_squish_command function would reduce the repetitiveness in the code. I will make that change.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
RenaL27: Since farm ids are unique identifiers that differ across different accounts and environments, if a developer uses a farm id, it would only be specific to their environment and wouldn't work for other developers running the e2e tests. By using names, we can make sure that the tests can run with different developers where there is a farm and queue with the name in the config.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
RenaL27: That's true, I guess it doesn't need to have a macOS storage profile but it may be useful to have a test that uses one to verify that is included in the job parameters.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
RenaL27: I will delete the None default values since these parameters are pretty much required.

test/squish/suite_nuke_submitter/shared/scripts/cleanup_helpers.py
RenaL27: True, I think this is a better way to delete these output files.

test/squish/suite_nuke_submitter/shared/scripts/names.py
RenaL27: This is a Squish generated file that contains object identifiers for UI elements in the Nuke application and the Deadline Cloud submitter. These identifiers are used in the squish_helpers and squish gui testing.

test/squish/suite_nuke_submitter/shared/scripts/squish_helpers.py
RenaL27: If the user didn't install Nuke, Squish will fail and return an a startApplication failed error with the reason why. If squish was not installed correctly, it would fail because the squish import will not be recognized.

test/squish/suite_nuke_submitter/shared/scripts/squish_helpers.py
RenaL27: No, these number coordinates are not based on pixel count. They are relative coordinates within the UI element being clicked. Because the coordinates are relative to the UI element, and not the screen, it still works when the UI scales.

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -0,0 +1,136 @@
...
+    # Get the queue ID from configuration
+    queue_id = api_helpers.get_queue_id_by_name(farm_id)
+    assert queue_id is not None, "Queue ID not found"
+
+    # Get the most recent job ID from the queue
+    latest_job_id = api_helpers.get_latest_job_id(farm_id, queue_id)
+
+    # Verify the job ID from Squish matches the latest job in the queue
+    if job_id == latest_job_id:
+        print("Job ID from Squish execution matches the latest job ID.")
```
eherozhao, line 108: do we have a logging system in the nuke package? If so, we should reuse logger instead of print, which can give us a better logging level.

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+# mypy: disable-error-code="attr-defined"
```
eherozhao, line 2: out of curiousity: why do we need this?

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
```diff
@@ -0,0 +1,56 @@
...
+# submitting a job to Deadline Cloud, and then closing Nuke.
+#
+# This script is executed by the Squish test runner when the "basic_workflow_gui" test case is selected.
+# It is triggered using the run_squish_command function in run_squish_test.py in the test_basic_workflow test.
+#
+# The main() function is the entry point that Squish automatically calls when executing this test script.
+
+
+def main():
+    squish_helpers.launch_nuke()
```
eherozhao, line 21: q: do we need to have any try catch for these? I am not familiar with squish but I am wondering if nuke fails to be launched, will we just fail it or retry the launch?

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
```diff
@@ -0,0 +1,56 @@
...
+
+    file_path_edit = squish.waitForObject(names.script_to_open_FilePathEdit)
+    nuke_asset_path = os.environ.get("NUKE_ASSET_ROOT")
+    if not nuke_asset_path:
+        test.fatal("NUKE_ASSET_ROOT environment variable not set")
+        return
+
+    script_path = os.path.join(
+        nuke_asset_path,
+        "nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/scripts/nukeSubmitter_v02_nuke_default_one_write_node.nk",
```
eherozhao: nit: this can be put to a const variable.

eherozhao (commented):
more to think about:
1. I see there are couple places we are using hardcoded nuke version 16. When we release a new version, this will fail and we need to update the version. Is it possible to have a way to support all versions we want? not just hardcoded one.
2. It will be good to include a specific file to introduce how this tests were set up and how the operators should run the test in GitHub for any contributors to run the test on their own.

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
RenaL27: We need this because the squish, test, and name modules are provided by the Squish testing framework, which mypy doesn't have type information for, so disabling the error code attr-defined will help avoid "has no attribute" errors.

test/squish/suite_nuke_submitter/basic_workflow_gui/test.py
RenaL27: If Nuke fails to launch, the test will fail and Squish will print the error message of why. For example, if the floating license timed out, the test will throw and error saying "Couldn't get license from server: Connection refused
This is an UNLICENSED version of Squish. Please install your license key provided by The Qt Company Ltd. For assistance, use https://www.qt.io/contact-us#sales-contact"

test/squish/suite_nuke_submitter/run_squish_test.py
RenaL27: There does exist a logging system in the nuke package but the nuke package mainly uses print statements in its code for output and status reporting.

RenaL27:
> more to think about:
> 
> 1. I see there are couple places we are using hardcoded nuke version 16. When we release a new version, this will fail and we need to update the version. Is it possible to have a way to support all versions we want? not just hardcoded one.

Yes, it is possible to support all the versions we want as long as the Squish application uses the same QT version as the Nuke's QT version. Nuke 16 uses QT 6.5 and further updates versions of Nuke would still work with Squish as long as the QT version remains 6.5. Otherwise, the developer can download the appropriate Squish application that supports a different version of QT that's the same as the updated QT version that Nuke uses.

test/squish/SQUISH_README.md
```diff
@@ -0,0 +1,61 @@
...
+/path/to/DeadlineCloudForNukeSubmitter
+```
+You can also modify the envvars file:
+```sh
+NUKE_PATH=/Users/user/DeadlineCloudForNukeSubmitter
+```
+### Set Required Environment Variables
+The following environment variables are required for running the tests:
+```sh
+export AWS_PROFILE=<ada profile name>
```
jaikish: just `<profile name>`. `ada` is an internal tool

test/squish/SQUISH_README.md
```diff
@@ -0,0 +1,61 @@
...
+
+- An AWS default profile (used for authentication)
+- A farm named "Nuke Submitter Squish Farm"
+- A queue named "Nuke Submitter Squish Automation Queue"
+- Three storage profiles named "Linux Storage Profile", "Windows Storage Profile", and "macOS Storage Profile"
+- Fleet instance market type of On-Demand instance
+
+## Available Tests
+### Basic Workflow Test
+The `basic_workflow` test provides comprehensive end-to-end validation of the basic Nuke submitter workflow. This test consists of two main parts: First, the basic_workflow_gui Squish test automates the UI interaction by launching Nuke, opening a pre-configured script file with a write node, configuring AWS profile and resources, submitting a job to Deadline Cloud, and closing Nuke. Second, the test performs backend validation by verifying the job was successfully submitted to the correct queue, monitoring job completion, downloading the rendered output files, comparing them against reference images to ensure visual accuracy, and handling the cleanup of resourcse.
```
jaikish: Can you add some more details
 
1. How a successful test's response looks like vs unsuccessful one here? 
2. Where to check the different images when test fails
3. If there is an intended change to the UI, suggest updating the reference images at  `"/nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/images/expected",`

test/squish/SQUISH_README.md
```diff
@@ -0,0 +1,61 @@
+# Nuke Submitter E2E Testing with Squish
+
+Nuke and Squish require a license. If you have a Squish and Nuke license, please follow the guide below to run the tests.
+
+## Prerequisites
+### Install Nuke
+
+Download and install Nuke 16.0v1. The current tests have been validated on macOS 15.5. Support for Windows and Linux will be added after macOS support is established.
```
jaikish, line 8: Can we make the test assert the platform as first step and fail fast with an error like `"deadline nuke squish tests are only supported on macOS currently"`

jaikish:
Great first PR. Keep going!

test/squish/suite_nuke_submitter/run_squish_test.py
eherozhao: I can understand it is out of the scope of your project but I would like us to consider about this logging improvement. Might be worth to evaluate how much effort it needs if we can introduce the basic logging configurations and firstly use them in your testing. Then we can gradually update nuke package with this logging config.

RenaL27: Yes, it would be good to have this logging improvement but this PR mainly seeks to add basic Squish support. I think that adding logging can be considered in future PRs or as a stretch goal.

test/squish/SQUISH_README.md
```diff
@@ -0,0 +1,81 @@
+# Nuke Submitter E2E Testing with Squish
```
MarinMaksutaj, line 1: nit: We should probably update https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/DEVELOPMENT.md as well to mention something about the new Squish tests.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -0,0 +1,165 @@
+# -*- coding: utf-8 -*-
```
MarinMaksutaj: Non blocking: please for all py files created put our copyright message `# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.`

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -0,0 +1,140 @@
+import os
```
MarinMaksutaj, line 2: Overall good job on the code structure. Looks very organized!

test/squish/SQUISH_README.md
RenaL27: Discussed offline. Can make this change on another PR.

jaikish (approved):

MarinMaksutaj (approved):

PR #229, opened 2025-06-30
feat: Add Custom Settings and Write Node E2E tests

RenaL27:
The framework currently implements the Basic Job Submission test case, ensuring reliable end-to-end testing of the core submission workflow. This PR will introduce a Custom Settings Job Submission test case, Write Node selection tests, and multiple error handling tests. The E2E tests currently only supports Mac. Support for other operating systems will come soon.
The change was tested by running `hatch run squish:test`, which triggers the E2E integration tests in the `run_squish_test.py`file.
Since these changes are focused on the test automation framework and don't modify the core job bundle functionality, running the Job Bundle Output Tests is not required. The changes are isolated to the test framework implementation.
```
```

Yes, the changes are through clear function and parameter documentation, logging, error messages, and more.
No.
----

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -39,9 +39,13 @@ test = "pytest --no-cov {args:test/*/integ}"
 [envs.e2e.scripts]
 test = "pytest --no-cov {args:test/*/e2e}"
 
+[envs.squish]
+env-vars = { PYTHONPATH = "{env:PYTHONPATH:}:test/squish/suite_nuke_submitter" }
+
 [envs.squish.scripts]
-deps = "pip install numpy"
-test = "pytest run_squish_test.py::test_basic_workflow"
+deps = "pip install numpy pytest pytest-xdist"
+test = "pytest --override-ini addopts= {args:test/squish/suite_nuke_submitter/run_squish_test.py} -v -n 0"
+
 
 [envs.release]
 detached = true
--- a/test/squish/suite_nuke_submitter/custom_settings_gui/test.py
+++ b/test/squish/suite_nuke_submitter/custom_settings_gui/test.py
@@ -10,7 +10,7 @@ GUI test for submitting a Nuke job with custom settings to AWS Deadline Cloud.
 This test verifies that users can successfully modify and submit jobs with custom settings:
 This test configures custom job settings like the job name, job description, priority, max retries
 per task, max failed tasks, and continue on error. This test submits a job with these custom settings
-and ensures that the test ensures that all custom settings are properly applied to the submitted job
+and ensures that all custom settings are properly applied to the submitted job
 and that the job submission process works correctly with modified parameters.
 """
 
--- a/test/squish/suite_nuke_submitter/custom_settings_gui/test.py
+++ b/test/squish/suite_nuke_submitter/custom_settings_gui/test.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # mypy: disable-error-code="attr-defined"
 import names
 import squish_helpers
--- a/test/squish/suite_nuke_submitter/custom_settings_gui/test.py
+++ b/test/squish/suite_nuke_submitter/custom_settings_gui/test.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# mypy: disable-error-code="attr-defined"
+import names
+import squish_helpers
+import squish
+
+"""
+GUI test for submitting a Nuke job with custom settings to AWS Deadline Cloud.
+
+This test verifies that users can successfully modify and submit jobs with custom settings:
+This test configures custom job settings like the job name, job description, priority, max retries
+per task, max failed tasks, and continue on error. This test submits a job with these custom settings
+and ensures that the test ensures that all custom settings are properly applied to the submitted job
+and that the job submission process works correctly with modified parameters.
+"""
+
+
+def main():
+    squish_helpers.launch_nuke()
+    squish.setWindowState(
+        squish.waitForObject(names.nukeMainWindow_Foundry_UI_DockMainWindow),
+        squish.WindowState.Maximize,
+    )
+    squish_helpers.open_nuke_script(
+        "nuke_submitter_v02_nuke_default_test_samples",
+        "nukeSubmitter_v02_nuke_default_one_write_node.nk",
+    )
+    squish.type(squish.waitForObject(names.nukeMainWindow_DAG_DAG_Window), "<Ctrl+S>")
+    squish_helpers.open_nuke_submitter_gui()
+    squish_helpers.set_job_name("Custom Setting Submission")
+    squish_helpers.set_job_description("This test verifies submission with modified settings")
+
+    squish_helpers.set_priority(75)
+    squish_helpers.set_max_retries(3)
+    squish_helpers.set_max_failed_tasks(10)
+
+    squish_helpers.configure_aws_profile()
+    squish_helpers.set_continue_on_error()
+    squish_helpers.submit_job()
+    squish_helpers.close_nuke()
--- a/test/squish/suite_nuke_submitter/run_squish_test.py
+++ b/test/squish/suite_nuke_submitter/run_squish_test.py
@@ -4,6 +4,7 @@ import subprocess
 import sys
 from shared.scripts import api_helpers, verification_helpers, cleanup_helpers
 import pathlib
+import pytest
 
 
 def run_squish_command(testcase_name, local=True):
@@ -101,6 +102,32 @@ def test_valid_environment_variables():
         print(f"✓ {var_name} is valid: {value}")
 
 
+@pytest.fixture
+def cleanup_render_outputs():
+    """Fixture that manages cleanup of rendered image outputs with manual trigger points"""
+    cleanup_queue = []
+
+    def register_cleanup(output_dir: str, base_name: str):
+        """Register a cleanup task"""
+        cleanup_queue.append((output_dir, base_name))
+
+    def execute_cleanup():
+        """Execute all registered cleanups"""
+        while cleanup_queue:
+            output_dir, base_name = cleanup_queue.pop(0)
+            try:
+                success, message = cleanup_helpers.cleanup_output_images(output_dir, base_name)
+                if not success:
+                    raise RuntimeError(f"Cleanup failed: {message}")
+            except Exception as e:
+                raise RuntimeError(f"Cleanup error: {str(e)}")
+
+    yield register_cleanup, execute_cleanup
+
+    # Final cleanup of any remaining items
+    execute_cleanup()
+
+
 def test_invalid_conda_env():
     check_platform()
     success = run_squish_command("invalid_conda_gui")
@@ -114,8 +141,21 @@ def check_platform():
     ), "Deadline Nuke Squish Tests are only supported on macOS currently."
 
 
-def test_basic_workflow():
+def test_basic_workflow(cleanup_render_outputs):
     check_platform()
+
+    NUKE_ASSET_ROOT = os.environ.get("NUKE_ASSET_ROOT", "")
+    TEST_SAMPLES_DIR = (
+        f"{NUKE_ASSET_ROOT}/nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples"
+    )
+    EXPECTED_DIR = f"{TEST_SAMPLES_DIR}/images/expected"
+    OUTPUT_DIR = f"{TEST_SAMPLES_DIR}/images/output"
+    FRAME_RANGE = (101, 120)
+    RGB_TOLERANCE = 0.1
+
+    register_cleanup, _ = cleanup_render_outputs
+    register_cleanup(OUTPUT_DIR, "nukeTest_output_v01")
+
     success, job_id = run_squish_command("basic_workflow_gui")
     # Exit early if the Squish test failed
     if not success:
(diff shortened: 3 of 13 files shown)


test/squish/suite_nuke_submitter/custom_settings_gui/test.py
viknith, line 1: Nit: I recall asking this before but forgot why, but does each test need to be in it's own `test.py` file in its own unique folder? Why not just have files like  test_custom_settings_gui.py without it being in a subfolder?

test/squish/suite_nuke_submitter/custom_settings_gui/test.py
```diff
@@ -0,0 +1,40 @@
...
+import squish_helpers
+import squish
+
+"""
+GUI test for submitting a Nuke job with custom settings to AWS Deadline Cloud.
+
+This test verifies that users can successfully modify and submit jobs with custom settings:
+This test configures custom job settings like the job name, job description, priority, max retries
+per task, max failed tasks, and continue on error. This test submits a job with these custom settings
+and ensures that the test ensures that all custom settings are properly applied to the submitted job
```
viknith: Nit: "...and ensures that the test ensures that all custom settings" sounds weird

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -18,11 +19,8 @@ def run_squish_command(testcase_name, local=True):
             - success (bool): True if command executed successfully (return code 0)
             - job_id (str or None): Extracted job ID from output if found, None otherwise
     """
-    deadline_nuke_path = os.environ.get("DEADLINE_NUKE_PATH")
-    if not deadline_nuke_path:
-        print("DEADLINE_NUKE_PATH environment variable not set")
-        return False, None
```
viknith, line 24: If we're removing this, do we have any handling elsewhere where the tests will fail early if DEADLINE_NUKE_PATH is not set?

**Edit**: Ah, I just found the `test_valid_environment_variables` test, I'm guessing that's the one?

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -51,17 +49,17 @@ def run_squish_command(testcase_name, local=True):
...
         has_squish_failure = False
         failure_message = ""
         lines = result.stdout.split("\n")
-
+        job_ids = []
         for line in lines:
             # Look for job ID
             if "Found job ID:" in line:
-                job_id = line.split("Found job ID:")[1].strip()
+                job_ids.append(line.split("Found job ID:")[1].strip())
```
viknith, line 64: QQ for curiosity, in which situation would we get multiple job IDs printing?

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -73,24 +71,59 @@ def run_squish_command(testcase_name, local=True):
...
     except Exception as e:
         print(f"Error running squish command: {e}")
         return False
 
 
-def test_basic_workflow():
+def test_valid_environment_variables():
+    # Test that required environment variables are set and valid.
+    env_vars = {
+        "AWS_PROFILE": {"check_path": False},
```
viknith, line 88: Should AWS_PROFILE not be set?

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -100,34 +133,76 @@ def test_basic_workflow():
...
+        + "/nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/images/expected",
+        output_dir=os.environ.get("NUKE_ASSET_ROOT", "")
+        + "/nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/images/output",
+        base_name="nukeTest_output_OCIO_v01",
+        start_frame=101,
+        end_frame=120,
+        rgb_diff_tolerance=0.1,
+    )
+
+    cleanup_success, cleanup_message = cleanup_helpers.cleanup_output_images(
```
viknith: If any of the assertions above fail, will cleanup get skipped? Ideally you'd want this to run at the very end, maybe as a cleanup fixture in pytest for example.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -84,10 +63,38 @@ def get_latest_job(farm_id, queue_id, job_id):
...
+
+    Returns:
+        bool: True if the job exists in the queue, False otherwise
+    """
+    try:
+        # Get all jobs in the queue
+        jobs = list_jobs(farmId=farm_id, queueId=queue_id)
+
+        # Check if the jobs list is empty
+        if not jobs or "jobs" not in jobs or not jobs["jobs"]:
```
viknith: Nit: you can do `if jobs.get("jobs", None)` or something like that. Same for any other instances of this, but I don't think there were that many.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -37,42 +37,21 @@ def get_queue_id_by_name(farm_id):
     return queue_id
 
 
-def get_latest_job_id(farm_id, queue_id):
-    """Get latest job from queue"""
```
viknith, line 41: QQ: why did we go from getting the latest job ID vs not? Was that approach no longer needed?

hatch.toml
```diff
@@ -39,9 +39,13 @@ test = "pytest --no-cov {args:test/*/integ}"
 [envs.e2e.scripts]
 test = "pytest --no-cov {args:test/*/e2e}"
 
+[envs.squish]
+env-vars = { PYTHONPATH = "{env:PYTHONPATH:}:test/squish/suite_nuke_submitter" }
+
 [envs.squish.scripts]
-deps = "pip install numpy"
-test = "pytest run_squish_test.py::test_basic_workflow"
+deps = "pip install numpy pytest pytest-xdist"
```
viknith, line 46: Does this deps call need to be done before test? If so, are the dev docs updated to show this?

RenaL27: Yes, this needs to be done before the test. I will update the squish readme to reflect this!

test/squish/suite_nuke_submitter/custom_settings_gui/test.py
RenaL27: In Squish, each folder is a separate test case in the Squish test suite. Squish treats each test.py as one test case. So each test case folder is only meant to hold one test case.

test/squish/suite_nuke_submitter/run_squish_test.py
RenaL27: Yes, `test_valid_environment_variables` tests to make sure that all the environment variables needed are set.

test/squish/suite_nuke_submitter/run_squish_test.py
RenaL27: In the write nodes selection test, the test does 2 job submissions, one with one write node selected and the other with all nodes option selected. Since this is done in one test case, 2 job IDs would need to be returned for testing purposes.

test/squish/suite_nuke_submitter/run_squish_test.py
RenaL27: The AWS_PROFILE environment variable is set so that the squish helper `configure_aws_profile` knows what profile to select in the dropdown menu in the submitter GUI.

test/squish/suite_nuke_submitter/run_squish_test.py
RenaL27: A cleanup fixture sounds like a good way to make sure that cleanup doesn't get skipped. I will add this to my test cases.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
RenaL27: This is because it is cleaner and less error prone if we get the job based off the id returned from the squish gui test rather than assuming that the latest job ID is the job we are currently testing.

hatch.toml
```diff
@@ -39,9 +39,13 @@ test = "pytest --no-cov {args:test/*/integ}"
...
+test = "pytest --override-ini addopts= {args:test/squish/suite_nuke_submitter/run_squish_test.py} -v -n 0"
```
MarinMaksutaj, line 47: Do we want verbose by default here?

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -73,24 +72,98 @@ def run_squish_command(testcase_name, local=True):
...
         print(f"Error running squish command: {e}")
         return False
 
 
-def test_basic_workflow():
+def test_valid_environment_variables():
+    # Test that required environment variables are set and valid.
+    env_vars = {
+        "AWS_PROFILE": {"check_path": False},
+        "DEADLINE_NUKE_PATH": {"check_path": True},
```
MarinMaksutaj, line 89: Just curious why did we decide for a new DEADLINE_NUKE_PATH, vs having just NUKE_PATH?

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -73,24 +72,98 @@ def run_squish_command(testcase_name, local=True):
...
+    check_platform()
+
+    NUKE_ASSET_ROOT = os.environ.get("NUKE_ASSET_ROOT", "")
+    TEST_SAMPLES_DIR = (
+        f"{NUKE_ASSET_ROOT}/nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples"
+    )
+    EXPECTED_DIR = f"{TEST_SAMPLES_DIR}/images/expected"
+    OUTPUT_DIR = f"{TEST_SAMPLES_DIR}/images/output"
+    FRAME_RANGE = (101, 120)
+    RGB_TOLERANCE = 0.1
```
MarinMaksutaj: non blocker, any way we could have some of these hard coded versions in a constants.py of some sort?

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -100,42 +173,156 @@ def test_basic_workflow():
...
+    check_platform()
+
+    NUKE_ASSET_ROOT = os.environ.get("NUKE_ASSET_ROOT", "")
+    TEST_SAMPLES_DIR = (
+        f"{NUKE_ASSET_ROOT}/nuke_test_samples/nuke_submitter_v02_nuke_modified_test_samples"
     )
+    EXPECTED_DIR = f"{TEST_SAMPLES_DIR}/images/expected"
+    OUTPUT_DIR = f"{TEST_SAMPLES_DIR}/images/output"
+    FRAME_RANGE = (101, 120)
+    RGB_TOLERANCE = 0.1
```
MarinMaksutaj: Same qs as above, if we can reuse this values from a constants file it would be good, or a top level variable in this file.

hatch.toml
RenaL27: Yes, I think verbose by default is useful here because it displays the name of each test function along with its pass/fail status and provides more detailed information on test failures, which is useful for testing.

test/squish/suite_nuke_submitter/run_squish_test.py
RenaL27: The DEADLINE_NUKE_PATH is the path that leads to the deadline-cloud-for-nuke directory. On the other hand, NUKE_PATH is the path that leads to nuke executable or submitter. I think it is more clear to use DEADLINE_NUKE_PATH so the user doesn't confuse it for the nuke executable path.

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -37,42 +37,21 @@ def get_queue_id_by_name(farm_id):
...
-
-    print("\n=== Latest Job ID Information ===")
-    print(latest_job_id)
-    return latest_job_id
-
-
-def get_latest_job(farm_id, queue_id, job_id):
-    # Returns the job details if found, None otherwise
+def get_job(farm_id, queue_id, job_id, check_storage_profile=False):
+    # Returns the job details if found, None otherwise. This also servers to verify that the default farm/queue and the optional storage is selected.
```
jaikish: typo: serves

test/squish/suite_nuke_submitter/shared/scripts/api_helpers.py
```diff
@@ -84,10 +63,38 @@ def get_latest_job(farm_id, queue_id, job_id):
...
+
+        # Check if the jobs list is empty
+        if not jobs["jobs"]:
+            print(f"No jobs found in queue {queue_id}")
+            return False
+
+        # Check if the job_id exists in the list of jobs
+        job_ids = [job["jobId"] for job in jobs["jobs"]]
+
+        if job_id in job_ids:
```
jaikish: minor: can be simplified to

```
if any(job["jobId"] == job_id for job in jobs["jobs"]):
    print(f"Job {job_id} found in queue {queue_id}")
    return True
print(f"Job {job_id} not found in queue {queue_id}")
return False
```

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -100,42 +173,156 @@ def test_basic_workflow():
...
-        output_dir=os.environ.get("NUKE_ASSET_ROOT", "")
-        + "/nuke_test_samples/nuke_submitter_v02_nuke_default_test_samples/images/output",
+    register_cleanup, execute_cleanup = cleanup_render_outputs
+    success, job_ids = run_squish_command("write_node_gui")
+
+    # Verify test execution was successful
+    if not success:
+        assert success, "Squish command failed"
+    if len(job_ids) != 2:
+        assert len(job_ids) == 2, f"Expected exactly one job ID, but got {len(job_ids)}"
```
jaikish: The assert is for `2` and the message says `one job ID`. Intentional or a typo?

RenaL27: That was a typo. I will fix this!

jaikish (dismissed):

MarinMaksutaj (approved):

viknith (approved):

PR #235, opened 2025-07-08
feat: Add OCIO, job attachment tests, and default frame range integration tests

RenaL27:

The framework currently implements the Basic Job Submission test case, custom settings job submission test case, write node selection test cases, etc, ensuring reliable end-to-end testing of the core submission workflow. This PR will introduce OCIO color management, job attachments, and frame range test cases. The E2E tests currently only supports Mac. Support for other operating systems will come soon.

The change was tested by running hatch run squish:test, which triggers the E2E integration tests in the `run_squish_test.py` file.

```
```

Yes, this change was documented in the SQUISH_README markdown file. The test cases were given a description of how they function in this markdown file.
No
----

### DIFF
--- a/test/squish/SQUISH_README.md
+++ b/test/squish/SQUISH_README.md
@@ -52,28 +52,35 @@ The following Deadline Cloud resources are needed in order to run `tst_verify_se
 - Fleet instance market type of On-Demand instance
 
 ## Available Tests
-### Basic Workflow Test
-The `basic_workflow` test provides comprehensive end-to-end validation of the basic Nuke submitter workflow. This test consists of two main parts: First, the basic_workflow_gui Squish test automates the UI interaction by launching Nuke, opening a pre-configured script file with a write node, configuring AWS profile and resources, submitting a job to Deadline Cloud, and closing Nuke. Second, the test performs backend validation by verifying the job was successfully submitted to the correct queue, monitoring job completion, downloading the rendered output files, comparing them against reference images to ensure visual accuracy, and handling the cleanup of resource.
 
-### Custom Settings Submission Test
-The `custom_settings` test validates the customization capabilities of the Nuke submitter interface for Deadline Cloud jobs. This test automates the UI interaction by launching Nuke, opening a pre-configured script file, and accessing the submitter interface to configure various job parameters. It sets custom values for job metadata (name and description), performance settings (priority level 75, maximum 3 retries per task, maximum 10 failed tasks), and error handling options (continue on error). After configuring these parameters, the test submits the job and verifies that all custom settings are correctly applied in the submitted job configuration. This ensures that the submitter interface reliably handles non-default parameter configurations and maintains setting integrity throughout the submission process.
+### Common Test Flow
+All tests follow a standard pattern:
+- Launch Nuke and open a pre-configured script
+- Configure AWS profile and resources
+- Submit job to Deadline Cloud
+- Verify job submission and configuration
+- Download and validate rendered outputs
+- Clean up resources
 
-### Write Node Selection Tests
-The `write_node_selection` test provides comprehensive validation of the write node selection functionality in the Nuke submitter interface. This test executes two distinct submission scenarios: first submitting a job that targets a single write node ("Write1"), and then submitting another job that processes all write nodes in the script. For each scenario, the test automates the UI interaction by launching Nuke, configuring the AWS profile, selecting the appropriate write node option, and submitting the job. After submission, the test performs backend validation by verifying the jobs were created with correct write node parameters, downloading the rendered outputs, comparing them against reference images with RGB difference tolerance checks, and cleaning up the output files. This ensures that the submitter correctly handles both individual and batch write node selections, maintaining proper job configuration and output generation in both scenarios.
+### Test Categories
 
-### Job Attachments Tests
-The job attachments tests provide validation of both automatic and manual file attachment capabilities in the Nuke submitter interface. These tests consist of two main scenarios:
+#### Basic Workflow
+Validates the basic end-to-end submission workflow using default settings. Verifies job submission, rendering, and output generation with a simple write node configuration.
 
-The `auto_detected_attachments` test verifies that the submitter correctly identifies and includes all necessary files referenced within the Nuke script. This test automates the UI interaction by launching Nuke with a pre-configured script, submitting the job with default settings, and then performs backend validation to ensure all script-referenced files are properly detected and included in the job submission.
+#### Custom Settings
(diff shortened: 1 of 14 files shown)


test/squish/SQUISH_README.md
```diff
@@ -61,6 +61,21 @@ The `custom_settings` test validates the customization capabilities of the Nuke
...
 The `write_node_selection` test provides comprehensive validation of the write node selection functionality in the Nuke submitter interface. This test executes two distinct submission scenarios: first submitting a job that targets a single write node ("Write1"), and then submitting another job that processes all write nodes in the script. For each scenario, the test automates the UI interaction by launching Nuke, configuring the AWS profile, selecting the appropriate write node option, and submitting the job. After submission, the test performs backend validation by verifying the jobs were created with correct write node parameters, downloading the rendered outputs, comparing them against reference images with RGB difference tolerance checks, and cleaning up the output files. This ensures that the submitter correctly handles both individual and batch write node selections, maintaining proper job configuration and output generation in both scenarios.
 
+### Job Attachments Tests
+The job attachments tests provide validation of both automatic and manual file attachment capabilities in the Nuke submitter interface. These tests consist of two main scenarios:
```
viknith: nit: I'd recommend simplifying the descriptions of these tests to be much shorter, maybe a one-line description per category of test to make it more readable. If there's anything common that each test does, you can cover that in one section and then have a second section about the unique things each test does. Non-blocking

test/squish/suite_nuke_submitter/shared/scripts/verification_helpers.py
```diff
@@ -62,3 +67,130 @@ def verify_image_sequence_rgb_matches(
...
+    """
+    Compare two video sequences by extracting frames and comparing their RGB values.
+    Verifies that both videos have matching resolution, duration, and frame count,
+    while also checking that the content matches between corresponding frames.
+
+    Args:
+        expected_video: Path to the expected reference video
+        output_video: Path to the output video to compare
+        frame_interval: Sample every nth frame (default: 1, meaning check every frame)
+        rgb_diff_tolerance: Maximum allowed RGB difference (default: 0.1)
```
viknith, line 87: Super cool!

test/squish/suite_nuke_submitter/shared/scripts/verification_helpers.py
```diff
@@ -62,3 +67,130 @@ def verify_image_sequence_rgb_matches(
...
+        if ember_path not in all_paths:
+            missing_files.append(ember_path)
+
+    # Check individual files
+    individual_files = [
+        "intro_to_compositing_test_samples/images/input/sh009_RAW_v001_1200.exr",
+        "intro_to_compositing_test_samples/images/input/sh009_STN_monster_BTY_v001_1200.exr",
+        "intro_to_compositing_test_samples/images/input/videoplayback.mov",
+        "intro_to_compositing_test_samples/scripts/Shot002.v01.001.nk",
+    ]
```
viknith: Are these hardcoded to a specific job submission? Would this need to be more generic or are we only testing with one video sequence? What if we had multiple video sequences?

test/squish/suite_nuke_submitter/suite.conf
```diff
@@ -2,6 +2,6 @@ AUT=Nuke16.0v1
 ENVVARS=envvars
 LANGUAGE=Python
 OBJECTMAPSTYLE=script
-TEST_CASES=basic_workflow_gui
+TEST_CASES=basic_workflow_gui custom_settings_gui default_frame_range_gui invalid_conda_gui manual_attachments_gui ocio_gui write_node_gui
```
viknith, line 5: Just curious, can this be modified such that it will try to run all test cases rather than specifying the ones to run?

test/squish/suite_nuke_submitter/shared/scripts/verification_helpers.py
```diff
@@ -62,3 +67,130 @@ def verify_image_sequence_rgb_matches(
...
+        output_video: Path to the output video to compare
+        frame_interval: Sample every nth frame (default: 1, meaning check every frame)
+        rgb_diff_tolerance: Maximum allowed RGB difference (default: 0.1)
+    """
+    # Open both videos
+    ref_cap = cv2.VideoCapture(expected_video)
+    out_cap = cv2.VideoCapture(output_video)
+
+    if not ref_cap.isOpened() or not out_cap.isOpened():
+        raise ValueError("Failed to open one or both video files")
```
viknith, line 94: Same question for anywhere where we raise Errors or exceptions. Will they be caught at the very top of the test execution such that we can have a finally block that executes cleanup?

test/squish/suite_nuke_submitter/suite.conf
RenaL27: Yes, we do not need to modify the test case part of the suite.conf for all the tests to all run. When we are running the end to end tests, the test cases call the corresponding gui test using squishrunner. So the E2E test cases will run all these test cases mentioned in the suite.conf even if it is not listed.

test/squish/SQUISH_README.md
RenaL27: Yes, that would make it more digestible and clear for the reader. I will make that change!

test/squish/suite_nuke_submitter/shared/scripts/verification_helpers.py
RenaL27: This is hardcoded to a specific job submission. Since we are only testing with one video sequence, I think it is best to not make it a function and instead include it in the test case. I will make this change!

jaikish (dismissed):

test/squish/suite_nuke_submitter/shared/scripts/verification_helpers.py
RenaL27: Yes, the cleanup will still execute even when errors are raised. This is because the cleanup_render_outputs fixture is a pytest fixture, and pytest guarantees that fixture teardown will run even if the test fails with an exception. The cleanup fixture will make sure that assets and files generated by the test will be cleaned up.

test/squish/suite_nuke_submitter/suite.conf
viknith: But everyone time someone adds a test, they'll need to update this .conf file right? Any way we can avoid that extra step?

RenaL27: Nope, if someone adds a test, they do not have to specifically specify it in the .conf file because Squish automatically detects all the test cases located in the test suite directory structure. Adding it in the .conf file here can be seen as an extra optional step.

viknith (approved):

jaikish (approved):

PR #238, opened 2025-07-10
feat: Add frame range tests, rewrote QCombobox interactions

RenaL27:
The framework currently implements the basic job submission, custom settings job submission, write node selection, job attachments, and OCIO test cases, ensuring reliable end-to-end testing of the core submission workflow. This PR will introduce frame range test cases. The E2E tests currently only supports Mac. Support for other operating systems will come soon.
The change was tested by running hatch run squish:test, which triggers the E2E integration tests in the run_squish_test.py file. Tests are optimized using pytest-xdist with -n 4 for parallel execution across 4 worker processes, while filelock ensures GUI operations remain sequential to prevent window overlap. This allows the non-GUI operations (like API calls and verification) to run in parallel while maintaining proper GUI synchronization, where a test's GUI portion must complete before the next one begins. This approach has significantly reduced test execution time from 40 minutes to approximately 10 minutes.

```
```

Yes, this change was documented in the SQUISH_README markdown file. The test cases were given a description of how they function in this markdown file.
No.
----

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -43,7 +43,7 @@ test = "pytest --no-cov {args:test/*/e2e}"
 env-vars = { PYTHONPATH = "{env:PYTHONPATH:}:test/squish/suite_nuke_submitter" }
 
 [envs.squish.scripts]
-deps = "pip install numpy pytest pytest-xdist opencv-python"
+deps = "pip install numpy pytest pytest-xdist opencv-python filelock"
 test = "pytest --override-ini addopts= {args:test/squish/suite_nuke_submitter/run_squish_test.py} -v -n 4"
 
 
--- a/hatch.toml
+++ b/hatch.toml
@@ -44,7 +44,7 @@ env-vars = { PYTHONPATH = "{env:PYTHONPATH:}:test/squish/suite_nuke_submitter" }
 
 [envs.squish.scripts]
 deps = "pip install numpy pytest pytest-xdist opencv-python"
-test = "pytest --override-ini addopts= {args:test/squish/suite_nuke_submitter/run_squish_test.py} -v -n 0"
+test = "pytest --override-ini addopts= {args:test/squish/suite_nuke_submitter/run_squish_test.py} -v -n 4"
 
 
 [envs.release]
--- a/test/squish/suite_nuke_submitter/custom_frame_range_gui/test.py
+++ b/test/squish/suite_nuke_submitter/custom_frame_range_gui/test.py
@@ -1,0 +1,52 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# mypy: disable-error-code="attr-defined"
+import names
+import squish_helpers
+import squish
+import test
+
+"""
+GUI test for submitting a Nuke job with custom frame range settings to AWS Deadline Cloud.
+
+This test verifies that users can successfully override and submit jobs with custom frame ranges:
+This test enables the frame range override option and configures a specific frame range (1-10) 
+instead of using the default range from the Nuke script. This test submits a job with the custom 
+frame range and ensures that the override is properly applied to the submitted job.
+"""
+
+
+def main():
+    squish_helpers.launch_nuke()
+    squish.setWindowState(
+        squish.waitForObject(names.nukeMainWindow_Foundry_UI_DockMainWindow),
+        squish.WindowState.Maximize,
+    )
+    squish_helpers.open_nuke_script("intro_to_compositing_test_samples", "Shot002.v01.001.nk")
+    squish.snooze(3)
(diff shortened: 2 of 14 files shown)


hatch.toml
```diff
@@ -44,7 +44,7 @@ env-vars = { PYTHONPATH = "{env:PYTHONPATH:}:test/squish/suite_nuke_submitter" }
 
 [envs.squish.scripts]
 deps = "pip install numpy pytest pytest-xdist opencv-python"
-test = "pytest --override-ini addopts= {args:test/squish/suite_nuke_submitter/run_squish_test.py} -v -n 0"
+test = "pytest --override-ini addopts= {args:test/squish/suite_nuke_submitter/run_squish_test.py} -v -n 4"
```
jaikish, line 47: Does it open 4 windows in parallel and do the tests?

Can you update the PR description with the significance of `-n 4` and if you observed any improvement in test duration?

test/squish/suite_nuke_submitter/custom_frame_range_gui/test.py
```diff
@@ -0,0 +1,52 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# mypy: disable-error-code="attr-defined"
+import names
+import squish_helpers
+import squish
+import test
```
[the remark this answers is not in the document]
jaikish: We keep seeing this comment. Can this be fixed?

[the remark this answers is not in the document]
RenaL27: Since test is already imported in squish_helpers, I will create a test_log function to avoid importing test in this file and to get rid of this comment.

hatch.toml
RenaL27: No, it does not open 4 windows in parallel. It runs the test cases right after each other even if one test hasn't finished yet and only 4 test cases are running max at once, but the gui part doesn't happen in parallel because of my use of filelock. I will update the PR description to highlight the significance and it's improvements! Without `-n 4`, the tests took 40 minutes but after using `-n 4`, the tests took 10 minutes total.

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -475,6 +477,121 @@ def test_default_frame_range(cleanup_render_outputs):
...
+    )
+    register_cleanup(
+        TestConstants.get_output_img_dir(TestConstants.INTRO_COMPOSITION_TEST_SAMPLES_DIR),
+        "Shot002_v01_001",
+        render_settings_path,
+    )
+
+    success, job_id = run_squish_command("write_node_limit_gui")
+    if not success:
+        assert success, "Squish command failed"
```
viknith: This looks a bit unnecessary haha

test/squish/suite_nuke_submitter/run_squish_test.py
```diff
@@ -475,6 +477,121 @@ def test_default_frame_range(cleanup_render_outputs):
...
+    if len(job_id) != 1:
+        assert len(job_id) == 1, f"Expected exactly one job ID, but got {len(job_id)}"
+
+    farm_id = api_helpers.get_farm_id_by_name()
+    assert farm_id is not None, "Farm ID not found"
```
viknith: Instead of farm_id is not None, you could probably just do assert farm_id and if it exists it'll be a truthy value. Same for everywhere else this applies.

RenaL27: That is true. I will make that change!

test/squish/suite_nuke_submitter/run_squish_test.py
RenaL27: I will simplify this since the if block is not needed!

test/squish/suite_nuke_submitter/custom_frame_range_gui/test.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+
+
+def main():
+    squish_helpers.launch_nuke()
+    squish.setWindowState(
+        squish.waitForObject(names.nukeMainWindow_Foundry_UI_DockMainWindow),
+        squish.WindowState.Maximize,
+    )
+    squish_helpers.open_nuke_script("intro_to_compositing_test_samples", "Shot002.v01.001.nk")
```
MarinMaksutaj, line 23: nit: we might add more of these sample nk files in the future. Consider putting them somewhere as constants.

MarinMaksutaj (approved):

hatch.toml
jaikish: Nice!

jaikish (approved):

PR #240, opened 2025-07-15
fix: resolve NUKE_PATH duplication when individual installer called f…

MarinMaksutaj:
…rom combined installer

When pairing with our combined installer logic we would have a nested incorrect path

Simplified the path by using whatever gets passed by combined installer

Correct Nuke path for Mac

rebuild installer to verify path is correct

Documentation was changed accordingly to display nuke support for mac

No

### DIFF
--- a/install_builder/deadline-cloud-for-nuke.xml
+++ b/install_builder/deadline-cloud-for-nuke.xml
@@ -47,7 +47,7 @@
                 <setInstallerVariable name="nuke_installdir" value="${installdir}" />
             </actionList>
             <elseActionList>
-                <setInstallerVariable name="nuke_installdir" value="${installdir}/Submitters/Nuke" />
+                <setInstallerVariable name="nuke_installdir" value="${installdir}" />
             </elseActionList>
         </if>
 	</readyToInstallActionList>


install_builder/deadline-cloud-for-nuke.xml
```diff
@@ -47,7 +47,7 @@
                 <setInstallerVariable name="nuke_installdir" value="${installdir}" />
             </actionList>
             <elseActionList>
-                <setInstallerVariable name="nuke_installdir" value="${installdir}/Submitters/Nuke" />
+                <setInstallerVariable name="nuke_installdir" value="${installdir}" />
```
joel-wong-aws, line 50: I'm confused. `nuke_installdir` seems to just be a location for file storage. Why do we need to drop `/Submitters/Nuke`?

Also, if we keep this, the if-else condition is now redundant and can be simplified to just `<setInstallerVariable name="nuke_installdir" value="${installdir}" />`

eherozhao: same comment as Joel

MarinMaksutaj: Discussed offline, not needed for now so I'll close the PR.

PR #244, opened 2025-07-18
fix: All Views rendering bug to properly render multiple views instea…

MarinMaksutaj:
…d of repeating first view

There is currently a bug in the submitter where if a customer picks "All Views" we actually don't include any of the views in the list on the adaptor logic. 

Include all script views when "All Views" button is selected

Correctly render all views in deadline cloud render farms.

Added unit tests to verify logic works as intended, will also kick off a round of conda adaptor release for this change.
Built wheel files manually with the fixed adaptor, submitted a job that has the known issue, and verified that all views showed up in the output.

OS used: Mac
Nuke version: 16.0
SMF Linux
Conda Packages: nuke=16.* nuke-openjd=0.18.*

No need for documentation change, this is a bugfix

No

### DIFF
--- a/src/deadline/nuke_adaptor/NukeClient/nuke_handler.py
+++ b/src/deadline/nuke_adaptor/NukeClient/nuke_handler.py
@@ -191,9 +191,11 @@ class NukeHandler:
         views = data.get("views", [])
         NukeHandler._validate_non_empty_list_of_str(views, "views")
 
-        # The "All Views" value means to get all of them, and we do that per write node,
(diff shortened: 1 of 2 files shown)


RenaL27 (approved):

viknith (approved):

eherozhao (commented):
> Added unit tests to verify logic works as intended, will also kick off a round of conda adaptor release for this change.

is this enough for testing the fix? If you have submitted the job for any os settings, please list it here.

PR #254, opened 2025-08-18
fix: fix behavior of automatic input / output path detection

npmacl:

We have a couple bugs with how we are generating input / output paths for job attachments.

The first bug is just that we weren't considering the %v or %V expressions correctly, so projects that had multiple views weren't going to work properly. 

The other bug is that paths containing TCL expressions were not being evaluated correctly. In the case of something like `[tcl expression]/output/frame_####.png` this ends up working out OK, as we just give the parent directory of the file to job attachments. But if the path was `[tcl expression]/output/frame_####/frame.png`, then you end up with with the output directories being something like `evaluted_path/output/frame_0450` rather than having a directory for frame_0000, frame_0001, etc.

For input files, pass in context with each view and frame combination to getEvaluatedValue() to get all paths. For output files, just use the parent dir of the first filenode in the path containing an expression

fix bugs related to input / output filepaths for job attachments

added new unit tests, added new test cases, tested job bundle output tests.

tested with a couple nuke scenes containing multiple views. tested rendering locally, submitting to deadline, and confirmed downloaded output matched local results.

did not modify any files under `installer`

```


Timestamp: 2025-08-18T15:50:57.611129-07:00
Running job bundle output test: C:\Users\npmac\deadline-clients\deadline-cloud-for-nuke\job_bundle_output_tests\cwd-path

cwd-path
Test succeeded

Timestamp: 2025-08-18T15:50:58.365645-07:00
Running job bundle output test: C:\Users\npmac\deadline-clients\deadline-cloud-for-nuke\job_bundle_output_tests\frame

frame
Test succeeded

Timestamp: 2025-08-18T15:50:58.857421-07:00
Running job bundle output test: C:\Users\npmac\deadline-clients\deadline-cloud-for-nuke\job_bundle_output_tests\group-read

group-read
Test succeeded

Timestamp: 2025-08-18T15:50:59.333041-07:00
Running job bundle output test: C:\Users\npmac\deadline-clients\deadline-cloud-for-nuke\job_bundle_output_tests\multi-load-save

multi-load-save
Test succeeded

Timestamp: 2025-08-18T15:50:59.894408-07:00
Running job bundle output test: C:\Users\npmac\deadline-clients\deadline-cloud-for-nuke\job_bundle_output_tests\noise-saver

noise-saver
Test succeeded

Timestamp: 2025-08-18T15:51:00.462525-07:00
Running job bundle output test: C:\Users\npmac\deadline-clients\deadline-cloud-for-nuke\job_bundle_output_tests\ocio

ocio
Test succeeded

Timestamp: 2025-08-18T15:51:03.382003-07:00

```

no

no
----

### DIFF
--- a/src/deadline/nuke_submitter/assets.py
+++ b/src/deadline/nuke_submitter/assets.py
@@ -2,10 +2,10 @@
 
 from __future__ import annotations
 
-import os
 import re
-from collections.abc import Generator
-from os.path import commonpath, dirname, join, normpath, samefile
+import os
+from os.path import commonpath, dirname, join, normpath, samefile, isfile
+from dataclasses import dataclass
 from sys import platform
 import nuke
 
@@ -13,10 +13,18 @@ from deadline.client.job_bundle.submission import AssetReferences
 from deadline.client.exceptions import DeadlineOperationError
 from deadline.nuke_util import ocio as nuke_ocio
 
-FRAME_REGEX = re.compile(r"(#+)|%(\d*)d", re.IGNORECASE)
+FRAME_VIEW_EXPRESSION_REGEX = re.compile(r"(%(\d*)d)|(%[vV])", re.IGNORECASE)
 FILE_KNOB_CLASS = "File_Knob"
 NUKE_WRITE_NODE_CLASSES: set[str] = {"Write", "DeepWrite", "WriteGeo"}
-JOB_ID_REGEX = re.compile(r"^job-[0-9a-z]{32}$")
+
+
+@dataclass
+class IOPath:
+    path: str
+    is_file: bool
+
+    def __hash__(self):
+        return self.path.__hash__()
 
 
 def get_nuke_script_file() -> str:
@@ -40,7 +48,7 @@ def get_scene_asset_references() -> AssetReferences:
     nuke.tprint("Walking node graph to auto-detect input/output asset references...")
     asset_references = AssetReferences()
     script_file = get_nuke_script_file()
-    if not os.path.isfile(script_file):
+    if not isfile(script_file):
         raise DeadlineOperationError(
             "The Nuke Script is not saved to disk. Please save it before opening the submitter dialog."
         )
@@ -57,7 +65,7 @@ def get_scene_asset_references() -> AssetReferences:
             is_read_node = read_knob.value()
 
         if is_read_node or node.Class() not in NUKE_WRITE_NODE_CLASSES:
-            for filename in get_node_filenames(node):
+            for iopath in get_input_paths_for_filenode(node):
                 # if the filename is in the install dir, ignore it.
                 if node is nuke.root():
                     # Windows / Linux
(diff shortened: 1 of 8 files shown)


src/deadline/nuke_submitter/assets.py
```diff
@@ -2,21 +2,29 @@
...
+from dataclasses import dataclass
 from sys import platform
 import nuke
 
 from deadline.client.job_bundle.submission import AssetReferences
 from deadline.client.exceptions import DeadlineOperationError
 from deadline.nuke_util import ocio as nuke_ocio
 
-FRAME_REGEX = re.compile(r"(#+)|%(\d*)d", re.IGNORECASE)
+FRAME_VIEW_EXPRESSION_REGEX = re.compile(r"(%(\d*)d)|(%[vV])", re.IGNORECASE)
```
MarinMaksutaj: nit: Might be worth adding a comment explaining what this regex does

MarinMaksutaj (dismissed):

src/deadline/nuke_submitter/assets.py
eherozhao: +1

src/deadline/nuke_submitter/assets.py
```diff
@@ -67,25 +75,31 @@ def get_scene_asset_references() -> AssetReferences:
...
                         pass
                     else:
                         if samefile(install_path, common_file_path):
                             continue
-                if not os.path.isdir(filename):
-                    asset_references.input_filenames.add(filename)
+
+                if iopath.is_file:
+                    asset_references.input_filenames.add(iopath.path)
+                else:
```
eherozhao: q: are we sure this else condition always means iopath is a directory?

src/deadline/nuke_submitter/assets.py
```diff
@@ -121,42 +135,103 @@ def find_all_write_nodes() -> set:
...
+    out = set()
     for knob in node.allKnobs():
-        if knob.Class() == FILE_KNOB_CLASS and knob.value():
-            # If the knob value starts with a tcl expression, we evaluate it
-            if knob.value().startswith("["):
-                yield normpath(join(project_path, knob.getEvaluatedValue()))
+        if knob.Class() != FILE_KNOB_CLASS or not knob.value():
+            continue
+
+        # gets the file path, still containing %04d, %v or TCL expressions.
```
eherozhao, line 178: what is tcl?

eherozhao:
During the review, I feel the submitter doesn't have many logs statement. Will we consider about adding more logs for better troubleshooting?

src/deadline/nuke_submitter/assets.py
npmacl: Yes, the IOPath type is defined by us, and has a bool type for the is_file member variable.

this if / else block is actually not needed as get_input_paths_for_filenode will only return iopath type where is_file is true. I can take it out for clarity.

src/deadline/nuke_submitter/assets.py
npmacl: A scripting language Nuke has support for. https://www.nukepedia.com/reference/Tcl/

An example of how this might appear would be something like `[file dirname [value root.name]]/output/frame_####.exr`, with `[file dirname [value root.name]]` being evaluated as the path to the nuke project file.

src/deadline/nuke_submitter/assets.py
npmacl: Can do.

npmacl:
> During the review, I feel the submitter doesn't have many logs statement. Will we consider about adding more logs for better troubleshooting?

Yeah that probably makes sense. Added a few logs to the parsing logic for output directories.

MarinMaksutaj (approved):

eherozhao (approved):

PR #278, opened 2025-12-08
fix: fix handling of python expressions in fileknob paths, add warning dialog to allow users to bypass asset scanning errors.

npmacl:
1. the previous job attachment scan assumed expressions within square brackets were always TCL expression. Nuke also supports python expressions here.
2. not all users of the submitter actually need job attachments to be correctly identified, for example users who intend to mount their own filesystem. bugs preventing the job attachments from being identified can block these users from submitting, even though their job would work fine.

1. replaced logic for evaluating tcl expressions with calling nuke.filename(node). this function has the desired behavior of evaluating both tcl and python expressions, while leaving frame and view expressions alone.
2. wrapped asset reference scanning with try catch exceptions, and created a new warning dialog that appears if an exception is caught. users can choose to ignore the warning and submit anyway.

1. users whose nuke scripts contain python expressions in paths will not get an exception on submission.
2. in the case of another bug encountered or introduced in the asset reference scanning path, customers who do not rely on this feature will be work around the bug.

### How was this change tested
tested the following using both nuke 16.0 and 14.1:
- tested submitting a job, confirmed it rendered correctly.
- added a deliberate exception in the asset references scan code path. confirmed the warning dialog appeared and worked as expected. either cancelling the submission or ignoring the warning and submitting worked as intended.
- ran unit tests

Here is what the warning dialogue looks like:
asset_scan_warning

```
== 4 passed, 9 skipped in 7.97s ==
```

yes, although there appears to be a transient bug currently that causes the first test in alphabetical order to fail. the cause seems to be related to a warning popup related to nuke licenses we are getting. changing the order of the tests show that all the test cases have the correct behavior.

yes

no
----


src/deadline/nuke_submitter/assets.py
```diff
@@ -175,20 +195,16 @@ def get_output_paths_for_filenode(node) -> set[IOPath]:
...
+        filepath = nuke.filename(node)
 
         expression_match = FRAME_VIEW_EXPRESSION_REGEX.search(filepath)
         path_is_file = True
         if expression_match:
             # in the case of an expression for frames / views, we will used the parent directory
             # of the filenode containing the first expression
             nuke.tprint(
-                f"found printf style expression {expression_match.group(1)} starting at index {expression_match.start()} in path {filepath}"
+                f"found printf style expression {expression_match.group(0)} starting at index {expression_match.start()} in path {filepath}"
```
eherozhao, line 207: Can you help me understand the change from 1 to 0 here?

src/deadline/nuke_submitter/assets.py
```diff
@@ -175,20 +195,16 @@ def get_output_paths_for_filenode(node) -> set[IOPath]:
...
             continue
 
-        # gets the file path, still containing %04d, %v or TCL expressions.
-        # note #### syntax for frames will be converted to %04d
-        # note backslashes for windows paths are converted to forward slashes
-        filepath = knob.value()
-        # evaluate any tcl expressions in the path
-        filepath = evaluate_tcl_subexpressions(filepath)
+        # evaluate any tcl / python expressions in the path, while leaving view / frame number expressions intact
+        filepath = nuke.filename(node)
```
eherozhao, line 199: Do we still evaluate tcl experession here?

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -6,19 +6,20 @@
...
 
 import nuke
+import yaml  # type: ignore[import]
 from deadline.client.api import get_deadline_cloud_library_telemetry_client
 from deadline.client.job_bundle import deadline_yaml_dump
 from deadline.client.ui import gui_error_handler
 from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import (  # type: ignore
-    SubmitJobToDeadlineDialog,
     JobBundlePurpose,
+    SubmitJobToDeadlineDialog,
```
eherozhao, line 17: q: was this change from some formatter?

src/deadline/nuke_submitter/assets.py
npmacl: Yes. I tested the function with both python and TCL expressions and it had the desired behavior.

src/deadline/nuke_submitter/assets.py
npmacl: the previous value of 1 was incorrect here. in the case of matching `(%(\d*)d)`, group(1) would be `(\d*)` rather than the whole expression. in the case of (%v)|(#+) it would just be None. switching to group(0) gets the desired group

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
npmacl: yes. the formatter had a lot of changes it wanted to make. I didn't want to obfuscate my change so I only included the changes to the files I was already modifying.

src/deadline/nuke_submitter/assets.py
eherozhao: Can we add these tests as code here?

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
eherozhao: I think the formatter you ran was "hatch fmt". right?

eherozhao (approved):

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
npmacl: yes

src/deadline/nuke_submitter/assets.py
npmacl: our unit tests run by `hatch test` don't have access to nuke, they stub it instead. we will want to cover this case in our e2e tests though.

ishasat (approved):

PR #281, opened 2025-12-24
feat: add copycat adaptor script

npmacl:

We want to add support for the copycat feature in deadline. Copycat needs to be run differently than normal nuke, so we had to either update the existing adaptor or use a new one. We have some other refactoring we want to do with the adaptor at some point anyway, so adding a simple adaptor script seemed to make the most sense for now.

added simple adaptor script for copycat.

none yet. I am merging the pieces for copycat support piecemeal

1. created and ran the unit tests.
2. I have a copycat job bundle that uses this adaptor in one of the branches in my fork. I tested submitting a job using that template and this adaptor, and it all seemed to work correctly.

```
== 130 passed, 8 skipped in 12.22s ==
```

didn't bother for this change, the submitter isn't impacted at all.

```
```

no

----


src/deadline/nuke_adaptor/copycat_adaptor.py
```diff
@@ -0,0 +1,143 @@
...
+import logging
+import pathlib
+import re
+import subprocess
+import sys
+import threading
+import time
+from typing import Dict, List, Optional
+
+error_regex = re.compile(r".*ERROR: (.+)")
```
eherozhao: It seems there's a error regex variable in original adaptor and it covers more options: https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/src/deadline/nuke_adaptor/NukeAdaptor/adaptor.py#L163

src/deadline/nuke_adaptor/copycat_adaptor.py
```diff
@@ -0,0 +1,143 @@
...
+step_complete_regex = re.compile(r"\[Step:(\d+)/(\d+)\].*")
+
+_current_percent_done = 0
+
+
+def _report_progress(current_step, total_steps):
```
eherozhao: can you add a docstring for the method to help the other folks understand the code?

eherozhao: do we have typing for parameters in this package?

src/deadline/nuke_adaptor/copycat_adaptor.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+_error_encountered = False
+
+
+# returning message rather than printing to make this unit testable
+def report_openjd_messages(line) -> Optional[str]:
+    global _error_encountered
+    if (match := error_regex.match(line)) is not None:
+        # error message is captured in group(1)
```
eherozhao: where was this group(1) defined? I think you are referring to regex group but the comment might not help that much.

src/deadline/nuke_adaptor/copycat_adaptor.py
```diff
@@ -0,0 +1,143 @@
...
+        for path in (rule["source_path"], rule["destination_path"])
+    )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        prog="NukeCopyCatAdaptor",
+        description=(
+            "Wrapper around executing CopyCat nodes in Nuke. Handle path mapping from job attachments"
+            " and emmision of OpenJD status messages to stdout"
```
eherozhao: "emmision" typo?

src/deadline/nuke_adaptor/copycat_adaptor.py
```diff
@@ -0,0 +1,143 @@
...
+    for line in iter(stream.readline, ""):
+        line_stripped = line.rstrip()
+        if msg := report_openjd_messages(line_stripped):
+            # msg will be the str of the openjd message if there is something to print,
+            # None otherwise
+            logger.info(msg)
+        logger.info(f"{stream_name}: {line_stripped}")
+
+
+def get_nuke_remap_string(path_mapping_rules: List[Dict[str, str]]) -> str:
```
eherozhao, line 75: curious about how current adaptor doing path mapping, can you point it to me?

eherozhao: since it looks like we always replace windows path to linux way instead of vice versa, is there a reason why?

pyproject.toml
```diff
@@ -166,7 +166,7 @@ source = [
 
 [tool.coverage.report]
 show_missing = true
-fail_under = 72
```
eherozhao, line 169: Reducing this number doesn't look good, I think there's still some areas that we can add the unit test for the copycat adaptor, but not a blocker for this PR.

src/deadline/nuke_adaptor/copycat_adaptor.py
npmacl: I've only seen errors messages of this format with copycat, but yeah might as well cover the other possibilities as well.

src/deadline/nuke_adaptor/copycat_adaptor.py
npmacl: sure, can do

src/deadline/nuke_adaptor/copycat_adaptor.py
npmacl: I can try rewording this a bit.

src/deadline/nuke_adaptor/copycat_adaptor.py
npmacl: it's done here: https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/src/deadline/nuke_adaptor/NukeClient/nuke_client.py.

now that you mention it, it makes a lot of sense to put that into a common file and just reuse it.

interestingly `Path(path).as_posix()` as is done there will replace backslashes with forward slashes if the current OS is windows, but it will not work if the corrent OS is a posix system.

src/deadline/nuke_adaptor/copycat_adaptor.py
npmacl: yup meant to do emission here.

pyproject.toml
npmacl: my thoughts on this is that getting the uncovered lines in the adaptor, as it is written right now would require a bunch of finicky mocking of the sub-process related behavior. I don't think there is a way to mock that in a way that would actually be useful in determining if the program runs as expected, and it would just be a complex mess that is a liability to fix anytime someone touches this code.

I did have a test in an earlier version of this PR that I felt was actually very useful, which was to take advantage of the fact that the adaptor takes in a path to the nuke executable as an argument. I instead provided a path to a python script that would basically act as a stub for nuke, and ran the adaptor as a subprocess, in order to execute in a very realistic manner. I ran into some issues with that not working in the github testing environment despite working perfectly fine on both my windows and linux dev environments, but I think I have a couple ideas on how I might be able to fix that. Ill give that one another try.

src/deadline/nuke_adaptor/copycat_adaptor.py
npmacl: ah, I think I remembered why I didn't do this approach in the first place. that method I referenced relies on member functions from a parent class that is not within a class not defined in this package.

test/unit/deadline_adaptor_for_nuke/CopyCatAdaptor/test_copycat_adaptor.py
```diff
@@ -0,0 +1,34 @@
+import subprocess
```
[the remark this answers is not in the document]
eherozhao: Can we remove this import if it is not used? Also, I think we can add formatter/linter to recognize this during local dev.

test/unit/deadline_adaptor_for_nuke/CopyCatAdaptor/test_copycat_adaptor.py
```diff
@@ -0,0 +1,112 @@
...
+import sys
+
+import deadline.nuke_adaptor.copycat_adaptor as copycat_adaptor_module
```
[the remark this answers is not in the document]
eherozhao: this looks like the reasonble comment to me. We would need to use import .. from .. if it is possible.

test/unit/deadline_adaptor_for_nuke/CopyCatAdaptor/test_copycat_adaptor.py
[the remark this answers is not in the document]
npmacl: no, this comment is no longer accurate, it is used on line 72. this one is just out of date. I apologize that a lot of these comments are duplicated and are very annoying. last week when I tried deleting them the UI was bugged somehow and they couldn't be deleted. Ill try seeing if I can get rid of them now.

test/unit/deadline_adaptor_for_nuke/CopyCatAdaptor/test_copycat_adaptor.py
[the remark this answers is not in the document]
npmacl: this is used very specifically on line 117 to get the file path to the module. in all other cases the functions from that file are reference using the "import .. from .." syntax.

eherozhao (approved):

MarinMaksutaj (approved):

PR #283, opened 2026-01-06
feat: add copycat job template

npmacl:
running copycat requires using the new adaptor, and has different parameters we care about. this requires either significantly modifying the existing template to have two modes, or to just create a new job template.

created a new job template

none yet, until we merge in the submitter UI changes this job template isn't actually used.

I tested submitting this job to an SMF fleet, and confirmed that the job successfully performed copycat training as expected.

n/a

n/a

n/a. submitter is not impacted at all.

no

no
----


MarinMaksutaj (dismissed):

src/deadline/nuke_submitter/copycat_job_template.yaml
```diff
@@ -0,0 +1,67 @@
...
+  userInterface:
+    control: CHOOSE_INPUT_FILE
+    label: Nuke Script File
+    fileFilters:
+    - label: Nuke Script Files
+      patterns:
+      - '*.nk'
+    - label: All Files
+      patterns:
+      - '*'
```
eherozhao, line 17: I saw this was also used in regular nuke submitter template. How will this be used from a customer perspective?

src/deadline/nuke_submitter/copycat_job_template.yaml
```diff
@@ -0,0 +1,67 @@
...
+  description: The Nuke script file to render.
+
+- name: CopyCatNode
+  type: STRING
+  description: Which CopyCat node to train
```
eherozhao: Would "the name of CopyCat node to train" be be something more specific?

src/deadline/nuke_submitter/copycat_job_template.yaml
```diff
@@ -0,0 +1,67 @@
...
+- name: DataDir
+  type: PATH
+  objectType: DIRECTORY
+  dataFlow: INOUT
+  userInterface:
+    control: CHOOSE_DIRECTORY
+    label: Data Directory
+  description: The directory CopyCat data files are read from and written to.
+
+- name: CopyCatAdaptor
```
eherozhao, line 33: Can you help me understsand the purpose of this? Since this will be supported only in SMF, should this adaptor always be in the same path?

src/deadline/nuke_submitter/copycat_job_template.yaml
```diff
@@ -0,0 +1,67 @@
...
+  type: PATH
+  objectType: FILE
+  dataFlow: IN
+  default: .
+
+steps:
+
+- name: Train
+  parameterSpace:
```
eherozhao: If pramaeter space is empty, do we still need this as a part of step?

src/deadline/nuke_submitter/copycat_job_template.yaml
npmacl: 1. this would not be supported only in SMF, this would work on CMF as well. the customer would just need to make sure that the job environment has the nuke executable on the path.
2. since the file is included via job attachments, to be able to get it's path mapping via `{{Params.<parameter_name>}}` we need to add a parameter for it.

src/deadline/nuke_submitter/copycat_job_template.yaml
npmacl: probably not. let me try taking it out.

src/deadline/nuke_submitter/copycat_job_template.yaml
npmacl: sure, I can reword it.

src/deadline/nuke_submitter/copycat_job_template.yaml
npmacl: I wasn't familiar with this so checked the documentation: [link](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#22-jobpathparameterdefinition). apparently it is related to how the UI when performing GUI submit will function.

This option doesn't appear when using the submitter, instead it's value is populated [here](https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/c292eaa1ea96d8a4b356a3ca0453bde47b98219f/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py#L273). this might come into play if using `deadline bundle gui-submit`, although thats not working on my system currently.

eherozhao (approved):

MarinMaksutaj (approved):

PR #286, opened 2026-01-15
feat: add submitter for copycat.

npmacl:
We want to add support for CopyCat. To support this in a useful way, we need to add UI support for submitting a job to run copycat training.

Added a new button under Deadline Cloud menu. Users can select between submitting a Rendering Job, or a CopyCat job.

image

Clicking this option creates a job submission very similar to the flow for submitting a job to rendering. Job specific parameters in the UI are adjusted to accounted for different options.

image

Users will be able to submit copycat training jobs via the UI.

tested submitting a rendering job. confirmed it rendered correctly
tested submitting a copycat training job, confirmed the results were correct and that I could load the results into the DataDir and that the progress UI was updated correctly.

image
 
`141 passed, 8 skipped in 15.22s`

4 passed, 9 skipped in 108.22s (0:01:48)

yes, all 6 passed.

```
```

no
----

### DIFF
--- a/src/deadline/nuke_submitter/assets.py
+++ b/src/deadline/nuke_submitter/assets.py
@@ -22,6 +22,7 @@ from deadline.nuke_util import ocio as nuke_ocio
 FRAME_VIEW_EXPRESSION_REGEX = re.compile(r"(%(\d*)d)|(%v)|(#+)", re.IGNORECASE)
 FILE_KNOB_CLASS = "File_Knob"
 NUKE_WRITE_NODE_CLASSES: set[str] = {"Write", "DeepWrite", "WriteGeo"}
+COPYCAT_NODE_CLASS: str = "CopyCat"
 
 
 @dataclass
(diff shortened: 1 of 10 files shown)


src/deadline/nuke_submitter/assets.py
```diff
@@ -165,14 +170,28 @@ def get_input_paths_for_filenode(node) -> set[IOPath]:
     """Get all the file we will use as input for this node"""
 
     out = set()
+    project_path = get_project_path()
+    context = nuke.OutputContext()
+    views = nuke.views()
+
+    if node.Class() == COPYCAT_NODE_CLASS:
+        # CopyCat nodes can optionally reference a .cat file as initial weights.
```
MarinMaksutaj, line 178: A bit confused on this part, anywhere I could read more on it?

eherozhao (approved):

src/deadline/nuke_submitter/assets.py
npmacl: yeah, you can go to [this page](https://learn.foundry.com/nuke/content/reference_guide/air_nodes/copycat.html) and ctrl+f for "Initial Weights"

MarinMaksutaj (approved):

PR #291, opened 2026-01-22
chore: update changelog

npmacl:

I manually modified the 0.18.16 release to make it clear that the copycat feature was added in that version. The changelog file doesn't seem to pick that up though.

update the changelog.md file

changelog.md has better descriptions

n/A

n/A

n/A

n/A

n/A

no

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.18.16 (2026-01-20)
 
-
+Added support for training CopyCat nodes. Users are now able to select to submit a CopyCat training job. A CopyCat training job will perform CopyCat training on the render farm, rather than on the local system.
 
 ### Bug Fixes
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## 0.18.16 (2026-01-20)
 
-
+Added support for training CopyCat nodes. Users are now able to select to submit a CopyCat training job. A CopyCat training job will perform CopyCat training on the render farm, rather than on the local system.
```
viknith, line 3: Does this have a formal commit that works with this change, kinda like the commit messages below it?

npmacl: I don't have a specific commit to really use here- it was added over multiple commits, but they are all in 0.18.15. the only commit in 0.18.16 was a bugfix for something I discovered right before releasing.

viknith (approved):

eherozhao (approved):

PR #293, opened 2026-02-25
feat: Add task chunking support to Nuke submitter

rickrams:
Add opt-in task chunking using the OpenJD TASK_CHUNKING extension. When enabled, frames are grouped into contiguous chunks instead of being dispatched individually, reducing per-task overhead.  A future update could add non-contiguous chunking but for now this seems useful as is.

Changes:
- New chunked job template with CHUNK[INT] Frame parameter, ChunkSize, and TargetRuntimeSeconds job parameters
- RenderSettings: use_chunking, chunk_size, target_runtime_seconds fields (sticky, default off)
- Submitter selects chunked template when use_chunking is enabled
- UI: Enable task chunking checkbox with chunk size and target chunk duration controls

No adaptor changes needed - the existing NukeHandler.start_render() already handles the contiguous chunk frame range format (e.g. 1-10).

Don't have this setup so no.

No modified

Don't have this setup so no.

```
```

Yes

No

----

### DIFF
--- a/src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
+++ b/src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
@@ -3,6 +3,7 @@
 """
 UI widgets for the Scene Settings tab.
 """
+
 import os
 import nuke
 
@@ -98,6 +99,29 @@ class SceneSettingsWidget(QWidget):
         )
         lyt.addWidget(self.continue_on_error_check, 4, 0)
 
+        # Chunking controls
+        lyt.addWidget(QLabel("Chunk size"), 5, 0)
+        self.chunk_size_spin = QSpinBox(self, minimum=1, maximum=150)
+        self.chunk_size_spin.setValue(1)
+        self.chunk_size_spin.setToolTip(
+            "Number of frames to group into each chunk.\n"
+            "Use 1 for one frame per task (default).\n"
+            "Higher values reduce per-task overhead."
+        )
+        lyt.addWidget(self.chunk_size_spin, 5, 1, 1, -1)
+
+        lyt.addWidget(QLabel("Target chunk duration (seconds)"), 6, 0)
+        self.target_chunk_duration_spin = QSpinBox(self, minimum=0, maximum=86400)
+        self.target_chunk_duration_spin.setValue(0)
+        self.target_chunk_duration_spin.setToolTip(
+            "When set, the scheduler dynamically adjusts chunk sizes\n"
+            "based on observed runtimes of completed chunks, aiming\n"
+            "for this duration per chunk. Leave at 0 to use a fixed\n"
+            "chunk size for all chunks."
+        )
+        self.target_chunk_duration_spin.setSpecialValueText("Disabled")
+        lyt.addWidget(self.target_chunk_duration_spin, 6, 1, 1, -1)
+
     def _build_copycat_ui_options(self, lyt: QGridLayout):
         self.copycat_node_box = QComboBox(self)
         self._rebuild_copycat_node_drop_down()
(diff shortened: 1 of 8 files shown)


src/deadline/nuke_submitter/default_nuke_chunked_job_template.yaml
```diff
@@ -0,0 +1,162 @@
+specificationVersion: jobtemplate-2023-09
```
npmacl: don't see a need for creating a new job template. I think it would be better to update the existing job template to support chunking. when running without chunking we could just use  a chunk size of 1 with targetRuntimeSeconds set to 0.

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
```diff
@@ -98,6 +99,40 @@ def _build_render_ui_options(self, lyt: QGridLayout):
...
+        chunking_lyt.addWidget(self.chunk_size_spin, 0, 1)
+
+        chunking_lyt.addWidget(QLabel("Target chunk duration (seconds)"), 1, 0)
+        self.target_runtime_spin = QSpinBox(self, minimum=0, maximum=86400)
+        self.target_runtime_spin.setValue(0)
+        self.target_runtime_spin.setToolTip(
+            "When set, the scheduler dynamically adjusts chunk sizes after\n"
+            "the first chunks complete, aiming for this duration per chunk.\n"
+            "Useful when frame render times vary. Leave at 0 to use a fixed\n"
+            "chunk size for all chunks."
```
npmacl: "Useful when frame render times vary" seems doesn't seem like the right use-case for dynamic chunk sizing. this feature relies on the scheduler estimating how long each task will take based on the duration of previous tasks.

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
```diff
@@ -98,6 +99,40 @@ def _build_render_ui_options(self, lyt: QGridLayout):
         )
         lyt.addWidget(self.continue_on_error_check, 4, 0)
 
+        # Chunking controls
+        self.use_chunking_check = QCheckBox("Enable task chunking", self)
+        self.use_chunking_check.setToolTip(
+            "Group multiple frames into chunks to reduce application startup overhead per task"
+        )
+        lyt.addWidget(self.use_chunking_check, 5, 0)
```
npmacl: rather than  adding a check box for choosing whether to enable chunking, I think we are better just using the spinner for chunk size. customers who want no chunking can just use chunkSize of 1 with dynamic chunking disabled.

src/deadline/nuke_submitter/default_nuke_chunked_job_template.yaml
rickrams: Great idea - deleted the separate chunked template. Updated default_nuke_job_template.yaml in-place with TASK_CHUNKING extension and CHUNK[INT] Frame parameter. ChunkSize defaults to 1 (identical to old per-frame behavior).

src/deadline/nuke_submitter/ui/components/scene_settings_tab.py
rickrams: Good feedback, updated

npmacl:
tested running integration tests, they passed
```
Required test coverage of 69.0% reached. Total coverage: 70.83%
=================================================== slowest 5 durations ===================================================
2.68s call     test/unit/deadline_adaptor_for_nuke/CopyCatAdaptor/test_copycat_adaptor.py::test_execution_with_stubbed_nuke_fail
2.47s call     test/unit/deadline_adaptor_for_nuke/CopyCatAdaptor/test_copycat_adaptor.py::test_execution_with_stubbed_nuke_success
0.24s call     test/unit/deadline_adaptor_for_nuke/NukeAdaptor/test_adaptor.py::TestNukeAdaptor_on_cleanup::test_on_cleanup
0.24s call     test/unit/deadline_adaptor_for_nuke/NukeAdaptor/test_adaptor.py::TestNukeAdaptor_on_run::test_on_run
0.22s setup    test/unit/deadline_submitter_for_nuke/test_chunking.py::TestJobTemplate::test_template_has_task_chunking_extension
============================================= 155 passed, 8 skipped in 29.94s =============================================
```

npmacl:
did some manual testing, seems to work as expected.

the job bundle output tests fail though. Its doesn't seem to be a regression, the test code just has to be updated. I'll send you a zip with the results on slack, it should be pretty simple to update the tests accordingly.

npmacl:
We decided to move this to a different PR that includes a fix for the job output tests: https://github.com/aws-deadline/deadline-cloud-for-nuke/pull/294.

PR #294, opened 2026-03-26
feat: add task chunking

npmacl:
## Note:
This is a continuation of an earlier PR from @rickamas: https://github.com/aws-deadline/deadline-cloud-for-nuke/pull/293.
The changes are the same, except for adding some updates to the Job Bundle Output Tests.

## What is the change?
Add opt-in task chunking using the OpenJD TASK_CHUNKING extension. When enabled, frames are grouped into contiguous chunks instead of being dispatched individually, reducing per-task overhead. A future update could add non-contiguous chunking but for now this seems useful as is.

## Changes:

    New chunked job template with CHUNK[INT] Frame parameter, ChunkSize, and TargetRuntimeSeconds job parameters
    RenderSettings: use_chunking, chunk_size, target_runtime_seconds fields (sticky, default off)
    Submitter selects chunked template when use_chunking is enabled
    UI: Enable task chunking checkbox with chunk size and target chunk duration controls

No adaptor changes needed - the existing NukeHandler.start_render() already handles the contiguous chunk frame range format (e.g. 1-10).
## How was this change tested? 
- Local testing of some 500 frame scenes in Nuke 15.
- Tested submitting jobs with ChunkSize of 1 and TargetRuntimeSeconds  set to 0. Tested with ChunkSize of 1 and TargetRuntimeSeconds  set to a non-zero value.  tested ChunkSize >1 and TargetRuntimeSeconds  set to 0. Tested TargetRuntimeSeconds  > 0 and ChunkSize >1. All worked as expected.

## Please run the integration tests and paste the results below

`=== 155 passed, 8 skipped in 13.94s ===`

## If installer/ was modified or a file was added/removed from src/, then update the installer tests and post the test results below

No modified
## Did you run the "Job Bundle Output Tests"? If not, why not? If so, paste the test results here.

Ran the tests, all 6 passed.

## Was this change documented?

Yes
## Is this a breaking change?

No

### DIFF
--- a/job_bundle_output_tests/multi-load-save/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/multi-load-save/expected_job_bundle/template.yaml
@@ -1,4 +1,6 @@
 specificationVersion: jobtemplate-2023-09
+extensions:
+- TASK_CHUNKING
 name: multi-load-save.nk
 parameterDefinitions:
 - name: NukeScriptFile
@@ -20,6 +22,24 @@ parameterDefinitions:
   type: STRING
   description: The frames to render. E.g. 1-3,8,11-15
   minLength: 1
+- name: ChunkSize
+  type: INT
+  default: 1
+  minValue: 1
+  description: |
+    Number of frames per chunk. Use 1 for one frame per task (default). Higher values group frames into chunks to reduce per-task overhead.
+  userInterface:
+    control: SPIN_BOX
+    label: Chunk Size
+- name: TargetChunkDuration
+  type: INT
+  default: 0
+  minValue: 0
+  description: |
+    Optional target duration in seconds per chunk. When set, the scheduler dynamically adjusts chunk sizes based on observed runtimes of completed chunks. Set to 0 to use a fixed chunk size for all chunks.
+  userInterface:
+    control: SPIN_BOX
+    label: Target Chunk Duration (Seconds)
 - name: WriteNode
   type: STRING
   userInterface:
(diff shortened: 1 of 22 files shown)


ishasat (approved):

job_bundle_output_tests/multi-load-save/expected_job_bundle/template.yaml
```diff
@@ -20,6 +22,24 @@ parameterDefinitions:
   type: STRING
   description: The frames to render. E.g. 1-3,8,11-15
   minLength: 1
+- name: ChunkSize
+  type: INT
+  default: 1
+  minValue: 1
+  description: |
```
eng-josh, line 29: Nit: Should `maxValue: 150` be used here combined with the submitter UI of `maximum=150`?

job_bundle_output_tests/multi-load-save/expected_job_bundle/template.yaml
```diff
@@ -20,6 +22,24 @@ parameterDefinitions:
...
+    Number of frames per chunk. Use 1 for one frame per task (default). Higher values group frames into chunks to reduce per-task overhead.
+  userInterface:
+    control: SPIN_BOX
+    label: Chunk Size
+- name: TargetChunkDuration
+  type: INT
+  default: 0
+  minValue: 0
```
eng-josh, line 37: Nit: Should `maxValue: 86400` be used here combined with the submitter UI of `maximum=86400`?

eng-josh (approved):

PR #301, opened 2026-04-20
chore: process improvements

crowecawcaw:
Release process is a bit tedious.

- Improve pip and hatch caching with CI actions. (Caching was added in shared workflows already.)
- Add permissions to release note process so they can be LLM generated
- Add a single manual approval step instead of requiring manual approvals on every release step. Before merging, will require a new ManualReleaseGate environment and removing approvals for the release environment

Faster releases

Works in deadline-cloud repo

n/a

No

----

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -3,6 +3,9 @@ name: Code Quality
 on:
   pull_request:
     branches: [ mainline, release, 'patch_*' ]
+  push:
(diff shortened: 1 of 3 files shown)


baxeaz (approved):

npmacl (approved):

eng-josh:
We could previously do the update merge previously as in #291, but this behavior should not be used now as it locks up Semantic PR check.

PR #320, opened 2026-07-02
feat: implement NukeSubmitter on the unified BaseSubmitter

leon-li-inspire:
## Summary
Implements `NukeSubmitter` against the unified `BaseSubmitter` ABC (aws-deadline/deadline-cloud#1245), so Nuke can be driven through the shared submission contract in addition to the existing native GUI submitter. Following the Maya reference implementation (aws-deadline/deadline-cloud-for-maya#435), the module layout is **inverted**: the headless engine owns the builders and the GUI depends on it (rather than the GUI owning the builders and the unified API wrapping them).

## Design
- **`nuke_submitter/submitter.py`** — headless engine, no Qt:
  - `NukeSubmitter(BaseSubmitter)`: `get_settings` / `get_job_template` / `get_parameter_values` / `get_asset_references` (+ inherited `get_submission_context`).
  - The render builders `_get_job_template` / `_get_parameter_values` are `@staticmethod`s on the class, and consume a flat `NukeSubmitterSettings` directly (no native-settings translation step). Their helpers — `_get_write_node`, `_get_frame_list`, `_set_timeouts`, and the gizmo/OCIO template helpers — moved here too.
  - `NukeSubmitterSettings(BaseSubmitterSettings)` adds the render fields (write-node/view selection, proxy mode, continue-on-error, chunking, gizmos, timeouts). It is a **render-only** contract.
  - `get_settings` records the scene `frame_list` but leaves `override_frame_range` off, so the native fallback to a Write node's own `use_limit`/first/last range is preserved.
  - `get_asset_references` logs scan diagnostics (the GUI surfaces these in a dialog; the headless path has none) rather than silently returning an incomplete reference set.
- **`deadline_submitter_for_nuke.py`** — GUI layer, now depends on the engine:
  - Adds `_ui_settings_to_submitter_settings` (`name` → `job_name`) and drives `NukeSubmitter` for RENDER jobs, so the GUI and the headless API build the render bundle through **one code path**.
  - Dropped `_to_native_settings`.

## CopyCat training
CopyCat training is a GUI-only job type and cannot be expressed through the render-only unified settings, so its (small) template/parameter builders stay in the GUI module. `on_create_job_bundle_callback` dispatches **RENDER → `NukeSubmitter`**, **COPYCAT_TRAINING → the copycat path**. The timeout patcher (`_set_timeouts`) is shared between the two via a structural `Protocol`, so there's no duplication.

## Depends on
aws-deadline/deadline-cloud#1245 (base `BaseSubmitter` / `BaseSubmitterSettings`, re-exported from `deadline.client.api`).

## Testing
- `test_submitter.py` drives the engine: `get_settings`, `get_job_template`/`get_parameter_values` delegation + host-requirement injection, `get_asset_references` (incl. scan-diagnostic logging), and the moved render builders directly (frame-list/write-node fallback, queue-parameter conflict). The builder tests (`test_chunking` / `test_ocio` / `test_gizmos`) now import from `submitter.py`.
- Per the Maya reference, the submitter test modules import `BaseSubmitter` at load and are **not** skip-guarded, so they fail until a `deadline` release carrying the base is installed, and pass once it is. `submitter.py` is no longer omitted from coverage.
- Verified locally against a `deadline` exposing `BaseSubmitter`: **62 passed**, `ruff` clean. Note: CI's Python lint/test jobs are red until #1245 ships (the pinned PyPI `deadline` has no `BaseSubmitter`, so the imports and mypy's `attr-defined` check fail) — expected and self-resolving once the base is released.
- GUI publish → NukeSubmitter → CreateJob → **10/10 SUCCEEDED** render (10 EXRs) on a live farm (from the pre-inversion run; the render job bundle is unchanged). Rebased on current `mainline`.


src/deadline/nuke_submitter/submitter_api.py
```diff
@@ -0,0 +1,106 @@
...
+        """Convert SubmitterSettings to native SubmitterUISettings."""
+        native = SubmitterUISettings()
+        native.name = settings.name
+
+        render_settings = RenderSettings()
+        # _get_frame_list only honors frame_list when override_frame_range is set,
+        # otherwise it reads the scene's render globals. get_settings() populates
+        # frame_list from the scene but does not set the override flag, so enable it
+        # whenever we have a frame_list to submit.
+        render_settings.override_frame_range = settings.override_frame_range or bool(
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in d28ed25 — `get_settings` records the scene range in `frame_list` but leaves `override_frame_range=False`, and `_to_native_settings` now honors the flag as-is, so a Write node's own `use_limit` range is respected. Covered by `test_get_settings_populates_from_scene` + `test_to_native_settings_honors_override_flag`.

src/deadline/nuke_submitter/submitter.py
```diff
@@ -0,0 +1,106 @@
...
+        self,
+        settings: SubmitterSettings,
+        queue_parameters: list[dict[str, Any]],
+    ) -> list[dict[str, Any]]:
+        native_settings = self._to_native_settings(settings)
+        return _get_parameter_values(native_settings, queue_parameters)
+
+    def get_asset_references(self, settings: SubmitterSettings) -> dict[str, Any]:
+        outcome = get_scene_asset_references()
+        return outcome.asset_references.to_dict()
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in d28ed25 — `get_asset_references` now logs `high_level_exception` and each `failed_to_parse_nodes` entry at WARNING when the scan `encountered_exception()`, mirroring the UI warning dialog. Covered by `test_get_asset_references_logs_scan_diagnostics`.

test/unit/deadline_submitter_for_nuke/test_chunking.py
```diff
@@ -111,69 +111,62 @@ def test_template_passes_frame_range_to_adaptor(self, template):
...
-        )
+        from deadline.nuke_submitter.submitter import NukeSubmitter, NukeSubmitterSettings
 
         mock_ocio.is_OCIO_enabled.return_value = False
         mock_nuke.root.return_value = MagicMock()
         mock_nuke.root.return_value.frameRange.return_value = "1-100"
 
-        settings = SubmitterUISettings()
-        settings.jobtype_specific_settings = RenderSettings(
+        settings = NukeSubmitterSettings(
```
leon-li-inspire: Why don't we call create one NukeSubmitter and call `submitter.get_settings()` here, won't it get a different settings ?

src/deadline/nuke_submitter/submitter.py
```diff
@@ -0,0 +1,468 @@
...
+        # Record the scene frame range for callers that want it, but leave
+        # override_frame_range at its default (False). _get_frame_list honors
+        # that flag as-is, so the native fallback to a Write node's own
+        # use_limit/first/last (the UI default) is preserved unless a caller
+        # explicitly opts into overriding with the scene range.
+        settings.frame_list = str(nuke.root().frameRange())
+        settings.is_proxy_mode = nuke.root().proxy()
+        settings.input_filenames = [script_file] if script_file else []
+
+        return settings
```
moonyuet: Maybe we can do something like the pesudo code below?
```python
		script_file = get_nuke_script_file()
		return NukeSubmitterSettings(
		    job_name= Path(script_file).name if script_file else "Untitled",
		    project_path = get_project_path(),
		    frame_list = str(nuke.root().frameRange()),
		    is_proxy_mode = nuke.root().proxy(),
		    input_filenames = [script_file] if script_file else []
		
		)
```

src/deadline/nuke_submitter/submitter.py
```diff
@@ -0,0 +1,468 @@
...
+
+            # Add the environment to the end of the template's job environments
+            if "jobEnvironments" not in job_template:
+                job_template["jobEnvironments"] = []
+            job_template["jobEnvironments"].append(override_environment["environment"])
+
+        # Determine whether this is a movie render. If it is, we want to ensure that the entire Nuke
+        # evaluation is placed on one task.
+        write_node, write_node_name = _get_write_node(settings)
+        movie_render = "file_type" in write_node.knobs() and write_node["file_type"].value() in [
```
moonyuet, line 314: Isn't `file_type` always part of knobs for the write_node?

test/unit/deadline_submitter_for_nuke/test_chunking.py
moonyuet: You could check directly in Nuke I assumed if you got submitter with this codebase installed.
but I agree to use NukeSubmitter().get_settings() instead as it would use the updated setting and test with some edge cases.

src/deadline/nuke_submitter/submitter.py
leon-li-inspire: Done in c8aa776 — `get_settings` now builds `NukeSubmitterSettings` in a single constructor call as suggested. Behavior is unchanged: `override_frame_range` stays at its default (`False`) so the Write-node `use_limit` fallback in `_get_frame_list` is still honored. `test_get_settings_populates_from_scene` continues to pass.

src/deadline/nuke_submitter/submitter.py
leon-li-inspire: Not always — the guard is required. When the user leaves the WriteNode selection empty, `_get_write_node` returns `nuke.root()` (the whole comp), and the root node has no `file_type` knob, so `"file_type" in write_node.knobs()` is `False`. Only when a real Write node is selected is `file_type` present. This is unchanged from the pre-inversion code (it moved verbatim from `deadline_submitter_for_nuke.py`), so removing the membership check would raise a `KeyError` on the "All write nodes" path.

test/unit/deadline_submitter_for_nuke/test_chunking.py
leon-li-inspire: Done in c8aa776 — `test_parameter_values_include_chunking` now derives its settings from `NukeSubmitter().get_settings()` (mocking `nuke`/`get_nuke_script_file`/`get_project_path`) and then sets only the chunking fields, so it exercises the real settings the submitter produces rather than a hand-built dataclass. The other chunking cases keep an explicit `NukeSubmitterSettings()` where they are asserting defaults.

src/deadline/nuke_submitter/deadline_submitter_for_nuke.py
```diff
@@ -59,6 +53,11 @@
     CopyCatTrainingSettings,
     SubmitterUISettings,
 )
+from .submitter import (
```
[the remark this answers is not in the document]
leon-li-inspire: Confirmed — the pin guarantees these symbols, so no import guard is needed.

`BaseSubmitter` / `BaseSubmitterSettings` were added in deadline-cloud **0.60.2** (aws-deadline/deadline-cloud#1245). The pin is `deadline >= 0.60.2, < 0.61` (bumped up from `>= 0.60.1` in 632845d precisely because 0.60.1 lacks them). Verified in a clean venv:
- `0.60.1` → `hasattr(deadline.client.api, "BaseSubmitter")` is `False`
- `0.60.2` → `True`

Every version the range can resolve (0.60.2 and any future 0.60.x patch) therefore exports both symbols, so the top-level `from deadline.client.api import BaseSubmitter, BaseSubmitterSettings` cannot raise `ImportError` in-range — the CopyCat path is not taken down. (The test docstring note predates the pin bump; before 0.60.2 shipped, the import genuinely did fail, which is why the floor was raised.)

src/deadline/nuke_submitter/submitter.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if isinstance(settings, NukeSubmitterSettings):
+            render_settings.write_node_selection = settings.write_node_selection
+            render_settings.view_selection = settings.view_selection
+            render_settings.is_proxy_mode = settings.is_proxy_mode
+            render_settings.continue_on_error = settings.continue_on_error
+            render_settings.chunk_size = settings.chunk_size
+            render_settings.target_chunk_duration = settings.target_chunk_duration
+            native.include_gizmos_in_job_bundle = settings.include_gizmos_in_job_bundle
+            native.on_run_timeout_seconds = settings.on_run_timeout_seconds
```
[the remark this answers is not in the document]
leon-li-inspire: This concerns only the new headless API surface (a caller invoking `NukeSubmitter().get_job_template()` / `get_parameter_values()` directly). The **GUI submitter path is unaffected**: `on_create_job_bundle_callback` still runs the `timeouts_enabled` validation — including the zero-timeout rejection that raises `DeadlineOperationError` — before building the template, byte-identical to mainline (`deadline_submitter_for_nuke.py:243-255`). So there is no behavior change or regression for GUI users.

On the headless path, `_set_timeouts` writes the timeout values unconditionally, which matches the existing template contract (OpenJD spec 2023-09 requires timeouts hard-coded in the template). Honoring `timeouts_enabled` / validating zero-values on the API path would be a new feature rather than behavior-preservation, so I would prefer to track it separately unless you want it folded into this PR.

src/deadline/nuke_submitter/submitter_api.py
```diff
@@ -0,0 +1,129 @@
...
+                    "Failed to parse asset references from node '%s': %s",
+                    node_name,
+                    tb,
+                )
+        return outcome.asset_references.to_dict()
+
+    def _to_native_settings(self, settings: SubmitterSettings):
+        """Convert SubmitterSettings to native SubmitterUISettings."""
+        native = SubmitterUISettings()
+        native.name = settings.name
```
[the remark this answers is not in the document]
leon-li-inspire: Resolved by the inversion — this comment was about the old `_to_native_settings` adapter, which no longer exists. `description` is now inherited from `BaseSubmitterSettings` (present in 0.60.2) and consumed directly by `get_job_template` (`submitter.py:229-230`: `if settings.description: job_template["description"] = settings.description`). The GUI adapter also forwards it (`_ui_settings_to_submitter_settings`, `description=ui_settings.description`). Nothing is dropped on either path.

src/deadline/nuke_submitter/submitter.py
```diff
@@ -0,0 +1,466 @@
...
+                )
+
+        parameter_values.extend(
+            {"name": param["name"], "value": param["value"]} for param in queue_parameters
+        )
+
+        return parameter_values
+
+    def get_asset_references(self, settings: BaseSubmitterSettings) -> AssetReferences:
+        outcome = get_scene_asset_references()
```
MarinMaksutaj, line 454: Should this merge `settings.input_filenames / input_directories / output_directories` into the returned references? The `BaseSubmitterSettings` docstring says explicit inputs are "merged with scene-detected inputs in `get_asset_references`", but the scan result is returned as-is here — a headless caller's explicit entries (and the `input_filenames` that `get_settings` populates) are silently dropped.

src/deadline/nuke_submitter/submitter.py
```diff
@@ -0,0 +1,466 @@
...
+
+        return job_template
+
+    def get_parameter_values(
+        self,
+        settings: BaseSubmitterSettings,
+        queue_parameters: list[dict[str, Any]],
+    ) -> list[dict[str, Any]]:
+        settings = cast("NukeSubmitterSettings", settings)
+
```
MarinMaksutaj, line 349: Where do the inherited base fields (`priority, initial_status, max_failed_tasks_count, max_retries_per_task, max_worker_count, output_path`) get consumed on the unified path? On the GUI path the `deadline:*` parameters come from the shared dialog, but a headless caller via `get_submission_context` sets them with no effect. If that's a base-class question (and Maya behaves the same), fine to track separately — just want to confirm it's known.

src/deadline/nuke_submitter/submitter.py
```diff
@@ -0,0 +1,466 @@
...
+    else:
+        write_node = nuke.root()
+    return write_node, settings.write_node_selection
+
+
+def _get_frame_list(
+    settings: NukeSubmitterSettings,
+    write_node: Node,
+    write_node_name: Optional[str],
+) -> str:
```
MarinMaksutaj, line 168: This adds a fallback mainline didn't have: override on + empty frame list previously failed validation (Frames has `minLength: 1`), now it silently submits the scene/write-node range. Intentional for the GUI path too? A user who checks "Override frame range" and leaves it blank now gets a full-range render instead of an error — an explicit `DeadlineOperationError` there might be friendlier.

src/deadline/nuke_submitter/submitter.py
leon-li-inspire: Confirmed — this is a base-class-level concern, not specific to this PR, and Nuke mirrors Maya here. Neither the Maya submitter nor Nuke consumes `priority` / `initial_status` / `max_failed_tasks_count` / `max_retries_per_task` / `max_worker_count` / `output_path` from the settings object; on both DCCs the `deadline:*` job parameters are supplied through the queue parameters (the shared dialog on the GUI path, or `queue_parameters` / `initial_values` passed into `get_submission_context` on the headless path), not through these `BaseSubmitterSettings` fields.

So on the unified path these inherited fields are currently inert on both Nuke and Maya. Agreed it is worth wiring them (or trimming them from the base) — but that belongs at the `BaseSubmitter` layer so all DCCs behave consistently, so I would prefer to track it separately rather than solve it Nuke-only in this PR. Flagging as known.

src/deadline/nuke_submitter/submitter.py
leon-li-inspire: Good catch — fixed in 44e5c51. `get_asset_references` now unions the caller's explicit references from settings (`input_filenames` / `input_directories` / `output_directories`) with the scanned `AssetReferences`, matching the `BaseSubmitterSettings` field contract and the Maya submitter. This also ensures the scene file that `get_settings()` records in `input_filenames` is retained. New test `test_get_asset_references_merges_explicit_settings_references` covers it. The GUI path is unaffected (it builds its own `AssetReferences` from the submit dialog and does not call this method).

src/deadline/nuke_submitter/submitter.py
leon-li-inspire: Yes, intentional for both paths. The change guarantees the `Frames` parameter is never emitted empty (it has `minLength: 1`), so the previous "override on + blank" case — which failed server-side validation at CreateJob — now falls back to the scene / Write-node `use_limit` range and submits successfully. It's strictly a previously-failing case becoming a working one; every non-empty override value and the entire override-off path are unchanged.

We considered raising `DeadlineOperationError` on override-on-but-blank instead, but decided to keep the silent fallback: it's the more forgiving behavior, and the GUI already pre-fills the frame field with the scene range so a truly blank override is an edge case. Happy to revisit if you feel strongly that an explicit error is better UX.

leon-li-inspire: On reflection you're right — implemented in ce3bfe2. `override_frame_range` is an explicit opt-in to use the caller's range, so override-on with an empty `frame_list` is contradictory input; it now raises a clear `DeadlineOperationError` ("Frame range override is enabled but no frame range was provided. Enter a frame range or disable the override.") instead of silently rendering the full scene range.

The override-off path (the UI default) is unchanged — it still falls back to the scene / Write-node `use_limit` range, so `Frames` is never emitted empty (minLength: 1). Net effect vs. the three behaviors: this keeps the original bug fixed (no empty value) while replacing both the old cryptic server-side CreateJob rejection *and* my earlier silent fallback with an early, actionable client-side error that surfaces in the GUI via the submit dialog's error handler. Added tests for both the raise and the override-on-honors-value case. Thanks for pushing on this.

MarinMaksutaj (approved):

moonyuet:
It is published successfully from ayon-deadline-cloud in my side. 
image
[publish-report-260728-20-36.json](https://github.com/user-attachments/files/30462282/publish-report-260728-20-36.json)

Tested with manually publish also no specific issues found when submitting renders from Nuke.

homjason (approved):

PR #330, opened 2026-07-20
test: add xa11y-driven Nuke submitter UI tests (macOS)

eng-josh:
## What was the problem/requirement?

The GUI test suite for the Nuke submitter is built on Squish, which requires a commercial license, a matching Squish-for-Qt build per Nuke Qt version, and a real Deadline Cloud farm + AWS credentials to run. We are migrating GUI testing to [xa11y](https://github.com/xa11y/xa11y) driving the platform accessibility tree, with the mock Deadline backend from [deadline-cloud-test-fixtures](https://github.com/aws-deadline/deadline-cloud-test-fixtures) so tests run offline.

## What was the solution?

An opt-in suite at `test/nuke_submitter_ui/`, following the data-driven architecture proven in `deadline-cloud-for-cinema-4d`'s `test/integ_xa11y` (launcher/opener pattern originally modeled on `deadline-cloud`'s `test/blender_submitter_ui`):

- **Data-driven cases** — each case is a folder under `test_cases/` with an `input/scene.py` (executed inside GUI Nuke by the opener hook, so no extra license seat), an optional `input/configure.py` dialog hook, and committed `expected/job_bundle/` goldens. One parametrized runner (`test_submitter_ui.py`) exports a bundle per case and compares it structurally against the goldens, normalizing machine paths (`PATH_TO_BE_REPLACED`) and version-churning conda pins. Porting a Squish case = a case folder + one line in `_CASES`.
- **`_launcher.py`** — launches GUI Nuke hermetically: mock Deadline endpoint env, isolated `DEADLINE_CONFIG_FILE_PATH` (real `HOME` preserved so Foundry licensing works), and verified frontmost activation on macOS (the submitter is a `Qt.Tool` window, which macOS removes from the accessibility tree while Nuke is not frontmost).
- **`_opener/menu.py`** — a `NUKE_PATH` hook (inert without the suite's env vars) that runs the case's scene script, patches botocore's `management.` host prefix to reach the loopback mock, suppresses the update dialog, and opens the submitter via the same code path as the menu item.
- **`pages.py`** — `NukeSubmitterDialog` page object extending the fixtures' `SharedSubmitterDialog`, encoding macOS AX ground truth (canonical platform notes in its module docstring): combo popup items expose no working AX activation (physical click on the row, keyboard fallback); `QSpinBox` AX increment/decrement step by 10% of the range, so exact values are typed; a real click is required before typing because AX `focus()` does not make a `Qt.Tool` window key. Export confirmation is anchored on the success message-box body text; a queue-environment settle wait runs before driving the dialog. Most scene-settings widgets are anonymous in the AX tree, so accessors are positional for now — a follow-up accessible-naming pass on `scene_settings_tab.py` will make selectors robust.
- **Realistic mock latency** — the mock serves with 0.3 s per-response delay by default (`MOCK_DEADLINE_RESPONSE_DELAY_S` to override) so timing races aren't hidden.
- **Diagnostic modes** — `NUKE_SUBMITTER_UI_DIALOG_DUMP=1` dumps both settings tabs' accessibility trees for selector harvesting; `NUKE_SUBMITTER_UI_UPDATE_GOLDENS=1` regenerates a case's goldens from a verified run.
- **First case: `basic_workflow`** — port of the Squish `basic_workflow_gui` scenario (job name/description, write-node selection, chunk size, export) verified by golden comparison, `openjd check`, and mock-backend call assertions (nothing escapes to unmocked routes).

The suite is excluded from the default pytest run (`testpaths`) and skips when no Nuke installation is found. It requires a licensed GUI Nuke with this repo installed into Nuke's Python, accessibility permission for the terminal, and a hands-off machine while running (synthetic input goes to the focused window). Not wired into CI.

Run with:
```
python -m pytest test/nuke_submitter_ui -o addopts= -q
```

## What is the impact of this change?

Foundation of the Squish → xa11y migration; the remaining 8 Squish cases can be ported as case folders on this chassis. Squish stays in place until parity.

## How was this change tested?

- macOS 15.7.7, Nuke 16.0v7 (Qt 6.5.3), xa11y 0.11.0, deadline-cloud-test-fixtures 0.18.17: goldens bootstrapped via `UPDATE_GOLDENS=1` (normalization round-trip verified in the same run), then consecutive green runs against the committed goldens with 0.3 s mock latency (~15 s warm, ~35 s cold), fully offline.
- Committed goldens verified machine-independent (path placeholder applied; no usernames or absolute paths).
- `hatch run test`: 176 unit tests pass; default collection is unaffected. `hatch run lint` (ruff, black, mypy) clean.

## Was this change documented?

Yes: `test/nuke_submitter_ui/README.md` covers requirements, permissions, the canonical run command, the case layout, and the diagnostic env vars; `requirements-ui-testing.txt` pins the suite's dependencies; the accessibility-layer platform notes live canonically in the `pages.py` module docstring. Linux/Windows enablement is a follow-up.

## Is this a breaking change?

No. Test-only, opt-in.

----


eng-josh:
Addressed CI + scanner feedback in the amended commit:

- **mypy (the actual CI failure on the lint step):** `_opener/menu.py` env vars now default to `""` instead of `None` (fixes the two `str | None` errors), and `types-PyYAML` added to `requirements-testing.txt` for the `import yaml` stub error (same pattern as the existing `types-pywin32` pin).
- **CodeQL empty-except alerts:** all four now carry explanatory comments — each is a poll loop tolerating transient AX-tree churn (the dialog window drops out of the accessibility tree whenever Nuke loses frontmost).
- **CodeQL BaseException alerts (`_opener/menu.py`):** the inner handler is narrowed to `Exception`. The top-level `except BaseException` is intentional and mirrors this repo's own `src/menu.py`: a `menu.py` hook must never break Nuke startup, and the status file is the only failure-reporting channel out of the Nuke process. Happy to change if maintainers prefer otherwise.
- **DCO:** commit amended with sign-off.

test/nuke_submitter_ui/_launcher.py
```diff
@@ -0,0 +1,220 @@
...
+                    return binary
+    elif sys.platform.startswith("linux"):
+        for install_dir in sorted(Path("/usr/local").glob("Nuke*"), reverse=True):
+            binary = install_dir / install_dir.name.replace("v", ".").split(".", 2)[0]
+            matches = sorted(install_dir.glob("Nuke[0-9]*.[0-9]*"))
+            if matches:
+                return matches[0]
+            if binary.exists():
+                return binary
+    raise FileNotFoundError(
```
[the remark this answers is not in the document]
eng-josh: Good catch — fixed in 84d1564. The Linux branch now mirrors the macOS approach (first glob match that is a regular executable file) and the dead name-derived fallback is removed.

test/nuke_submitter_ui/_opener/menu.py
```diff
@@ -0,0 +1,95 @@
...
+   code path as the AWS Deadline menu item.
+3. Report success/failure to a status file polled by the launcher.
+"""
+
+import os
+import traceback
+
+_STATUS_FILE = os.environ.get("NUKE_SUBMITTER_UI_STATUS_FILE", "")
+_SCENE_FILE = os.environ.get("NUKE_SUBMITTER_UI_SCENE_FILE", "")
+_DELAY_MS = int(os.environ.get("NUKE_SUBMITTER_UI_OPEN_DELAY_MS", "5000"))
```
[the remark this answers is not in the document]
eng-josh: Fixed in 84d1564 — the int() parse is now wrapped in try/except ValueError falling back to the 5000ms default, keeping the module inert-and-safe at import as documented.

test/nuke_submitter_ui/_launcher.py
```diff
@@ -0,0 +1,221 @@
...
+    env["QT_LINUX_ACCESSIBILITY_ALWAYS_ON"] = "1"
+    (work_dir / "scene").mkdir(parents=True, exist_ok=True)
+    return env
+
+
+def _activate_app(pid: int, timeout: float = ACTIVATE_TIMEOUT) -> bool:
+    """Bring *pid*'s app to the foreground, verified via the AX bridge."""
+    if sys.platform != "darwin":
+        return True
+    from AppKit import (  # type: ignore[import-not-found]
```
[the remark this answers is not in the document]
eng-josh: Fixed in dfd6ebf — both profile vars are now restored symmetrically: each is copied from the real environment when defined and the mock redirection is dropped otherwise, so Windows (no HOME) no longer raises and doesn't inherit the fake profile dir either.

test/nuke_submitter_ui/utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+def log(message: str) -> None:
+    elapsed = time.monotonic() - _T0
+    timestamp = datetime.now().strftime("%H:%M:%S")
+    print(f"[nuke_submitter_ui {timestamp} +{elapsed:6.2f}s] {message}", flush=True)
+
+
+def repo_parent_path(any_path_in_repo: Path) -> str:
+    """The absolute path prefix that PATH_PLACEHOLDER stands for."""
+    return str(any_path_in_repo.resolve()).rsplit("deadline-cloud-for-nuke", 1)[0].rstrip("/\\")
```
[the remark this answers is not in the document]
eng-josh: Right conclusion — fixed in 6eebf23 by dropping the rstrip; the placeholder now always stands for everything before the literal `deadline-cloud-for-nuke` substring, separator included. One nuance on the diagnosis: the committed goldens' form doesn't come from the rstrip — this clone's directory is named `Bea-57276-deadline-cloud-for-nuke`, so the rsplit prefix ends in `-` and the rstrip was a no-op locally (also why the determinism runs passed). The failure you describe would have hit the first standard-named clone, where rsplit yields `/home/user/` and the rstrip truncates it. With the rstrip gone, both clone shapes round-trip and the committed goldens are already in the correct form.

test/nuke_submitter_ui/test_submitter_ui.py
```diff
@@ -0,0 +1,179 @@
...
+    write_deadline_config(
+        config_path,
+        farm_id=scenario.farm_id,
+        queue_id=scenario.queue_id,
+        job_history_dir=job_history_dir,
+    )
+    return config_path, job_history_dir
+
+
+def _assert_expected_mock_traffic(mock_backend) -> None:
```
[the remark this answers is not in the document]
eng-josh: Fixed in 7874075 — the inline block was a leftover from an edit that didn't apply cleanly; the test now calls _assert_expected_mock_traffic and the policy lives only in the helper.

karthikbekalp (dismissed):
LGTM. Thanks for working on Nuke tests right away!

test/nuke_submitter_ui/_launcher.py
```diff
@@ -0,0 +1,245 @@
...
+            for binary in sorted(install_dir.glob("Nuke*.app/Contents/MacOS/Nuke[0-9]*.[0-9]*")):
+                if binary.is_file() and os.access(binary, os.X_OK):
+                    return binary
+    elif sys.platform.startswith("linux"):
+        install_dirs = sorted(Path("/usr/local").glob("Nuke*"), key=_version_key, reverse=True)
+        for install_dir in install_dirs:
+            for binary in sorted(install_dir.glob("Nuke[0-9]*.[0-9]*")):
+                if binary.is_file() and os.access(binary, os.X_OK):
+                    return binary
+    raise FileNotFoundError(
```
npmacl, line 89: no case for windows?

eng-josh: Good catch — added in 4a8c55a: the Windows branch now discovers Nuke under Program Files (`Nuke<version>` dirs containing `Nuke<major>.<minor>.exe`), same newest-first version ordering as the macOS/Linux branches, with `NUKE_EXECUTABLE` still the universal override. One honest caveat noted in the commit: the suite itself has only been verified on macOS so far — the Windows discovery path is best-effort until we do the Windows enablement pass (UIA selectors differ, per the C4D suite's experience).

eng-josh: We will additionally be adding CI/CD harness for this in the future which will support Windows and Linux runs that will be able to verify the correctness of the path for Windows.

test/nuke_submitter_ui/_opener/menu.py
```diff
@@ -0,0 +1,112 @@
...
+        # raises AttributeError, which surfaces loudly via the status file.
+        from deadline.nuke_submitter import update_utils
+
+        update_utils._session_state.update_dismissed = True
+
+        from deadline.nuke_submitter import JobType, show_nuke_render_submitter
+        from PySide6.QtCore import qVersion
+
+        dialog = show_nuke_render_submitter(JobType.RENDER)
+        _write_status(
```
MarinMaksutaj, line 92: Should dialog is None be treated as an `ERROR` here? `show_nuke_render_submitter` can return `None`, and reporting OK in that case means the failure only shows up later as a 30s `wait_visible` timeout in `pages.py`.

test/nuke_submitter_ui/_launcher.py
```diff
@@ -0,0 +1,252 @@
...
+    env["QT_ACCESSIBILITY"] = "1"
+    env["QT_LINUX_ACCESSIBILITY_ALWAYS_ON"] = "1"
+    return env
+
+
+def _activate_app(pid: int, timeout: float = ACTIVATE_TIMEOUT) -> bool:
+    """Bring *pid*'s app to the foreground, verified via the AX bridge."""
+    if sys.platform != "darwin":
+        return True
+    from AppKit import (  # type: ignore[import-not-found]
```
MarinMaksutaj, line 137: Where does `AppKit` come from in a fresh venv? `xa11y 0.11.0` ships no dependencies (checked the wheel METADATA) and the fixtures package doesn't pull pyobjc either, so pip install -r `requirements-ui-testing.txt` on a clean macOS machine would hit an ImportError here mid-test. Should `pyobjc-framework-Cocoa` (with a macOS marker) be added to `requirements-ui-testing.txt`?

test/nuke_submitter_ui/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+)
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+
+from _launcher import find_nuke_executable  # noqa: E402
+
+
+def pytest_collection_modifyitems(config, items):
+    """These tests launch a licensed GUI Nuke — never run them in parallel."""
+    if config.pluginmanager.hasplugin("xdist") and config.getoption("numprocesses", 0):
```
MarinMaksutaj: Does this actually serialize? `xdist_group` is only honored under `--dist loadgroup`, and the repo's addopts set `-n auto` without a dist mode, so the default load scheduler ignores the marker. Moot with one case, but once more cases land, a plain pytest `test/nuke_submitter_ui `(without `-o addopts=`) would launch multiple GUI Nukes concurrently — synthetic input from parallel workers would land in each other's windows. Would it be safer to raise a `UsageError` here when xdist is active, pointing at the canonical command?

requirements-ui-testing.txt
```diff
@@ -0,0 +1,7 @@
+# Dependencies for the xa11y-driven GUI test suite in test/nuke_submitter_ui.
+# See test/nuke_submitter_ui/README.md for setup and usage.
+deadline-cloud-test-fixtures == 0.18.*
+xa11y == 0.11.*
+openjd-cli == 0.7.*
+PyYAML == 6.*
+pytest == 9.*
```
MarinMaksutaj: Should this mirror the conditional `pytest` pin from `requirements-testing.txt` (`9.1.* / 8.4.*` on python_version)? `pytest 9` needs `Python ≥3.10`, so this file fails to install on `3.9`, which the repo still supports. Alternatively, state a `≥3.10` floor for the suite in the `README`.

requirements-ui-testing.txt
```diff
@@ -0,0 +1,7 @@
+# Dependencies for the xa11y-driven GUI test suite in test/nuke_submitter_ui.
+# See test/nuke_submitter_ui/README.md for setup and usage.
+deadline-cloud-test-fixtures == 0.18.*
+xa11y == 0.11.*
+openjd-cli == 0.7.*
+PyYAML == 6.*
```
MarinMaksutaj: Is `PyYAML` used directly anywhere in the suite? I don't see a yaml import, and the fixtures package already declares `pyyaml>=6.0`, same question for the `types-PyYAML` addition in `requirements-testing.txt`. If they're leftovers from an earlier revision, both could be dropped.

test/nuke_submitter_ui/_opener/menu.py
eng-josh: Agreed — fixed in 9db7a08: a None return now writes ERROR to the status file (with the reason) and returns, so the launcher raises immediately with the opener status instead of the test failing 30s later on wait_visible. Verified with a live run.

test/nuke_submitter_ui/_launcher.py
eng-josh: Right, and good catch on the wheel METADATA — pyobjc was installed manually during development and never declared. Added `pyobjc-framework-Cocoa == 12.*; sys_platform == 'darwin'` to requirements-ui-testing.txt in 9db7a08.

test/nuke_submitter_ui/conftest.py
eng-josh: Confirmed — you're right that the marker was ineffective: xdist_group only serializes under --dist loadgroup, which the repo's `-n auto` addopts don't set (and my dev venv didn't even have xdist installed, which is why I never saw the hazard). Took your suggestion in 9db7a08: collection now raises UsageError pointing at the canonical command. Verified empirically with xdist installed: plain `pytest test/nuke_submitter_ui` fails fast, the canonical `-o addopts=` invocation collects normally.

requirements-ui-testing.txt
eng-josh: Mirrored the conditional pins from requirements-testing.txt (`9.0.*` on >=3.10, `8.4.*` below) in 9db7a08 rather than documenting a floor — the suite has no actual 3.10+ requirement.

requirements-ui-testing.txt
eng-josh: Correct on both — the suite's yaml import was removed in the golden-bundle restructure and I forgot to retire the deps. Dropped PyYAML from requirements-ui-testing.txt and types-PyYAML from requirements-testing.txt in 9db7a08. Verified mypy stays green without the stubs (src/'s yaml imports carry `# type: ignore[import]` per existing repo convention).

MarinMaksutaj (approved):

npmacl (approved):

PR #350, opened 2026-09-23
fix: pin the xa11y runner image instead of tracking macos-latest

andychoquette:
## What

`runs-on: macos-latest` → `runs-on: macos-15` in `integ_macos.yml`, with a comment recording why.

## Why

The xa11y suite has been red on mainline since 2026-09-21 — 8 of 9 cases, all identical:

```
AssertionError: Write-node selection failed: wanted 'Write1', combo shows 'All write nodes'
  test/nuke_submitter_ui/pages.py:339
```

`basic_workflow`, the only case that never calls `select_write_node`, still passes. Both the click path *and* the keyboard fallback fail, and per-case time went from the README's ~15–35 s to ~184 s — so the combo popup is timing out and being retried, not a locator that moved.

## Elimination

Everything else was constant between the last green run ([33095292798](https://github.com/aws-deadline/deadline-cloud-for-nuke/actions/runs/33095292798), 2026-08-27) and the first red one ([35652705485](https://github.com/aws-deadline/deadline-cloud-for-nuke/actions/runs/35652705485), 2026-09-21):

| Variable | Green | Red | Verdict |
|---|---|---|---|
| repo | `9f76c325c` | `d2221dd44` | one commit (#349), touches no UI or test file |
| `deadline-cloud-test-fixtures` | 0.18.20 | 0.18.22 | xa11y tree **byte-identical** |
| `deadline` | 0.60.5 | 0.60.7 | only `client/ui` change is a `~`-collapse fix in `path_widgets.py` |
| `xa11y` | 0.13.0 | 0.13.0 | unchanged |
| Nuke | 16.0v7 | 16.0v7 | unchanged |
| image major | macos-26 | macos-26 | unchanged |
| **image version** | **20260728.0273** | **20260907.0351** | **moved** |

The runner image is the only thing left. This suite drives real synthetic input through the macOS accessibility API, so an OS point release tightening that behaviour fits the symptom exactly: both input paths refused, timing out rather than mis-targeting.

## Honest caveats

- **This is elimination, not a reproduction.** I cannot run the suite outside CI, so the pin doubles as the test. Green ⇒ image confirmed. Still red ⇒ image exonerated, and the next step is re-harvesting the accessibility tree with `NUKE_SUBMITTER_UI_DIALOG_DUMP=1` on macos-26.
- **GitHub pins only the major label, not the patch version**, so this narrows the exposure rather than removing it.
- **`macos-15` is an older OS than users run.** That is the trade-off of using it as a bisect; the comment says to revisit macos-26 once the interaction is fixed for it.
- I deliberately did **not** enable `NUKE_SUBMITTER_UI_DIALOG_DUMP` here — it calls `pytest.fail()` by design and skips the export, so it would make every case fail and destroy the signal from this change.

## Follow-up worth considering

`timeout-minutes: 60` against a suite that now takes ~26 min (and one run that queued to ~2 h wall) is thinner headroom than it looks.


andychoquette:
## Confirmed by CI: 9 passed on macos-15

Ran the pin on `feature/ci-tests` as a single-variable test — same commit as the known-red run, only the image changed.

| Run | Image | Queued | Ran | Result |
|---|---|---|---|---|
| [35665070628](https://github.com/aws-deadline/deadline-cloud-for-nuke/actions/runs/35665070628) | macos-26 | 0.1 min | 28.1 min | 8 of 9 failed |
| [35910961278](https://github.com/aws-deadline/deadline-cloud-for-nuke/actions/runs/35910961278) | **macos-15** | 69.8 min | **5.4 min** | **9 passed** |

```
= 9 passed, 1 warning in 198.57s (0:03:18) =
```

So the elimination argument in the description now has a reproduction behind it. It is also a *fast* green — 5.4 min, in line with the historical healthy 6.0/6.2 min runs rather than the degraded 26–28 min. The retry-and-timeout behaviour disappears along with the failure, consistent with synthetic input being accepted rather than refused.

## On the 69.8 minute queue — not attributable to the pin

My first read was that `macos-15` is capacity-starved relative to `macos-latest`. That looks wrong. A concurrent diagnostic run on **`macos-latest`** has now also been queued 24+ minutes, against the same 0.0–0.1 min historical baseline, and one earlier `macos-latest` run queued 32 min. So macOS runner capacity is broadly constrained at the moment and the long queue is not a cost of pinning.

Worth noting `timeout-minutes` covers execution rather than queueing, so a long queue delays feedback but does not fail the job.

## Still worth doing separately

This pin restores a green signal today, but `macos-15` is an older OS than users run, so the interaction should eventually be fixed for macos-26. I have a diagnostic run in flight that dumps the write-node combo's accessibility tree on the failing image, which is what that fix would be built from.

MarinMaksutaj (approved):

crowecawcaw (approved):

